PR #12, opened 2019-04-12
implement authentication

Geal:
this PR adds the authentication method and authentication data parameters for the connection, and the corresponding `authenticate` method for `ConsumerBuilder`

### DIFF
--- a/client/src/connection.rs
+++ b/client/src/connection.rs
@@ -22,6 +22,12 @@ pub enum RequestKey {
     ProducerSend { producer_id: u64, sequence_id: u64 },
 }
 
+#[derive(Clone)]
+pub struct Authentication {
+  pub name: String,
+  pub data: Vec<u8>,
+}
+
 pub struct Receiver<S: Stream<Item=Message, Error=Error>> {
     inbound: S,
     outbound: mpsc::UnboundedSender<Message>,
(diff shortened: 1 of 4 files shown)


client/src/consumer.rs
```diff
@@ -26,6 +26,7 @@ impl<T: DeserializeOwned> Consumer<T> {
         sub_type: SubType,
         consumer_id: Option<u64>,
         consumer_name: Option<String>,
+        auth_data: Option<Authentication>,
```
stearnsc, line 29: line 38 here should read
```rust
Connection::new(addr, auth_data, executor.clone())
```

client/src/connection.rs
```diff
@@ -318,21 +324,23 @@ impl ConnectionSender {
 
 pub struct Connection {
     addr: String,
+    auth_data: Option<Authentication>,
```
stearnsc: It looks like this isn't used anywhere yet; were you planning on adding e.g. reconnect attempts or something where this is necessary? 

My instinct would be to avoid adding it to `Connection` until it's used to avoid the "field is never used" warning.

stearnsc (changes requested):
Looks good overall; 1 error and 1 warning that we should probably fix. Have you tested it against a running pulsar broker?

client/src/connection.rs
Geal: I added it first when I was experimenting, but it's not needed, yes

client/src/consumer.rs
Geal: sorry, I forgot to push that line

PR #16, opened 2019-04-19
api updates

Geal:
sorry if this PR mixes different things, I'm trying to set up the env for my use case (proxy in front of brokers, and partitioned topics), so I need proxy URLs and the partitioned lookup available.
If it's annoying, I'll try to separate in multiple PR next time

PR #17, opened 2019-04-23
Service discovery

Geal:
current work on adding service discovery support, based on #16

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -34,6 +34,8 @@ serde_json = "1.0.32"
 chrono = "0.4.6"
 futures-timer = "0.1.1"
 log = "0.4.6"
+trust-dns-resolver = "0.11"
+url = "1.7"
 
 [build-dependencies]
 prost-build = "0.4.0"
--- a/client/src/connection.rs
+++ b/client/src/connection.rs
@@ -247,9 +247,9 @@ impl ConnectionSender {
             .map_err(|_| ConnectionError::Disconnected)
     }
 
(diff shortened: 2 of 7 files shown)


Geal:
@stearnsc the PR is in a good state now, ready for review. There's a lot of features in there, some of which I'd like to extract elsewhere (like a connection manager that would be usable by SD, consumers and producers), but it's already usable

stearnsc:
This looks great!
My understanding of the new usage is the user:
(1) creates a `ServiceDiscovery`
(2) uses the `ServiceDiscovery` + relevant topic to obtain a `BrokerAddress`
(3) uses the `ServiceDiscovery` + `BrokerAddress` to obtain a `Connection`
(4) uses the `Connection` to create the `Producer` or `Consumer`
(5) uses the `Producer` or `Consumer` to send/receive messages.

That feels like a lot of steps to me; are you anticipating using helpers or similar to simplify the simplest "lookup topic, send messages" flows? Also I'm not sure I grok the `BrokerAddress` abstraction; is your vision that users would cache this somewhere in order to e.g. reconnect? What would users use this for other than immediately turn around and make a `Connection` with it?

(btw these questions are totally not blockers for this PR; I just want to get a sense for your vision)

stearnsc:
Ah sorry I read stuff out of order; lemme read your comments on https://github.com/wyyerd/pulsar-rs/issues/1 and see if it makes sense

Geal:
Also check out #20 for an interface idea that would make all those steps go away :)

stearnsc:
Got it, looks good!

Issue #20, opened 2019-04-25
new Client and ConnectionManager interfaces


Geal:
while working on the service discovery feature, I got an idea of how to make an API that's easy to use.

### Connection manager

Handling connections to a Pulsar cluster is a lot of work. Through a single client, we should be able to connect to multiple brokers, either directly, or through a proxy, in TCP plaintext or over TLS. A broker connection could be shared between multiple producers and consumers. Also, handling authentication.

I explored a way to handle it in the service discovery PR: a function that takes a broker address, and returns a `Future` of `Arc<Connection>`. If the connection exists, we get it immediately, otherwise we try to connect to that broker.

I'd like to extract that from the SD code, and have a `ConnectionManager` struct, that would be created from the following information:
- IP and port of either a broker or the proxy
- authentication data (if available)
- activate TLS or not

This connection manager would handle connecting to other broker, passing along authentication data, creating TLS connections, and answering queries for `Connection` object.
I'm not sure it should reconnect automatically when connections drop, because consumers using them would need to be created again, but it should allow reconnecting if asked to.

When used through SD, the connection manager would be automatically populated with the discovered brokers.
Consumers and producers could be created with connections coming from the manager (they already use a `Arc<Connection>` internally).

### pulsar::Client

This would be a wrapping object that can connect, perform lookups, create producers and consumers, while hiding the protocol details (the current `Connection`, `Consumer`, `Producer` and `ServiceDiscovery` APIs should still be available if people want to get a bit lower in the stack).

It would be created from connection info like the connection manager, and would have methods like these:
- `lookup_topic`/`lookup_partitioned_topic`
- `create_consumer`/`create_producer` from a broker address and a topic name
- `consume_topic`/`consume_partitioned_topic`, possibly creating one consumer per partitioned topic?
- `send_message` to send a single message to a topic

This `Client` interface would hide a lot of boilerplate to implement common use cases.

stearnsc:
I like this idea a lot;
>I'm not sure it should reconnect automatically when connections drop, because consumers using them would need to be created again, but it should allow reconnecting if asked to.

I think refactoring to allow transparently swapping out `Connection`s would be a good move eventually; since all that communication happens over channels, I don't foresee the mechanics being particularly difficult, but there are probably some issues to be worked out to ensure no data loss (from consumers) or duplicate writes (from producers). 

[...]

Issue #1, opened 2018-10-29
Correctly handle broker-discovery


stearnsc:
Currently we drop "LookupTopic" responses on the floor

Geal:
FYI I'm going to work on this now

Geal:
considering how lookup works (redirects, partitioned topics, proxys...), I think I'll make a `ServiceDiscovery` struct that will do the whole lookup.

There's another thing that we need to consider: the protocol is designed to handle connections to multiple brokers, either directly by TCP, or through a proxy.
A lookup with a redirect response requires contacting another broker.
Creating a producer or consumer for a partitioned topic requires contacting multiple brokers.
I am not sure yet, but I think a client cannot connect to multiple brokers through one proxy connection. Also the initial lookup would be done after a `Connect` command without proxy info, but then accessing the broker would need a new connection.

We could let the lib's user handle creating new connections as needed, but it might be useful to have a top `Client` object that can handle these tasks while hiding the multiple connections. Things like a consumer creation method for a partitioned topic that would do the lookup, create connections to each of the brokers if needed, create the consumers and return them.

Geal:
So I advanced a bit on service discovery in PR #17, and here's the kind of code I can write now:

```rust
    ServiceDiscovery::new(
      pulsar_addr.to_string(),
      Some(Authentication {
        name: auth_method_name.clone(),
        data: auth_data.clone(),
      }),
      runtime.executor()
    ).and_then(|mut sd| {
      sd.lookup_topic(topic)
    })
    .map(|v| {
      info!("topic lookup result: {:?}", v);
    })
```

For partitioned topics (I have to work a bit on error conversions again):

```rust
    ServiceDiscovery::new(
      pulsar_addr.to_string(),
      Some(Authentication {
        name: auth_method_name.clone(),
        data: auth_data.clone(),
      }),
      runtime.executor()
    ).and_then(|mut sd| {
     let mut sd2 = sd.clone();
      sd.lookup_partitioned_topic(topic)
        .map_err(|e| {
          error!("topic lookup error: {:?}", e);
          ServiceDiscoveryError::Dummy
        })
        .and_then(move |v| {
          info!("topic lookup result: {:?}", v);

          let res = v.iter().cloned().map(|(topic, broker_address)| {
            sd2.get_connection(&broker_address).map_err(|e| {
              error!("got error: {:?}", e);
              ServiceDiscoveryError::Dummy
            }).and_then(move |conn| {
[...]

PR #22, opened 2019-04-30
extract the connection manager from service discovery

Geal:
I got started on #20 and created a connection manager outside of the service discovery. It is designed the same way, a function that loops over queries coming from other futures. It gives access to connections either through the `BrokerAddress` information that comes from service discovery, or directly through a broker URL.
Maybe I should refactor the manager to only use broker URLs to index connections, but the interaction with proxys is not clear enough right now: lookups need to know if we have to go through a proxy or not, and the first connection is not proxying to a broker. I could add an argument to the manager's creation to indicate we expect to go through a proxy, that would simplify everything.

Now the example code I wrote in #1 looks like this:

```rust
    ConnectionManager::new(
        pulsar_addr.parse().unwrap(),
        Some(Authentication {
            name: auth_method_name.clone(),
            data: auth_data.clone(),
        }),
        runtime.executor(),
    )
    .from_err::<Error>()
    .and_then(|conn| {
        let manager = Arc::new(conn);
        ServiceDiscovery::with_manager(
            manager.clone(),
            runtime.executor(),
        ).into_future()
        .from_err()
        .and_then(|sd| {
            sd.lookup_partitioned_topic(topic)
                .from_err::<Error>()
                .map_err(|e| {
                    error!("topic lookup error: {:?}", e);
                    e
                })
        })
        .and_then(move |v| {
            info!("topic lookup result: {:?}", v);

            let res = v
                .iter()
                .cloned()
                .map(|(topic, broker_address)| {
                    manager
                        .get_connection(&broker_address)
                        .map_err(|e| {
                            error!("got error: {:?}", e);
                            e
                        })
                        .from_err()
                        .and_then(move |conn| {
                            Consumer::from_connection(
                                conn,
                                topic.to_string(),
                                "my_subscriber".to_string(),
                                SubType::Exclusive,
                                None,
                                None,
                                Box::new(|payload| deserialize(&payload.data)),
                                None,
                            )
                            .from_err()
                        })
                })
                .collect::<Vec<_>>();

            join_all(res)
        })
    })
    .and_then(|mut v| {
        info!("created {} consumers", v.len());

        let consumers = v
            .drain(..)
            .enumerate()
            .map(|(i, c)| {
                info!("will apply for_each on consumer {}", i);
                c.for_each(move |msg: Result<(Data, Ack), ConsumerError>| match msg {
                    Ok((data, ack)) => {
                        // process data
                        info!("consumer {} got data: {:?}", i, data);
                        ack.ack();
                        Ok(())
                    }
                    Err(e) => {
                        error!("consumer {} got an error: {:?}", i, e);
                        Ok(())
                        // return Err(_) to instead shutdown consumer
                    }
                })
            })
            .collect::<Vec<_>>();

        join_all(consumers).from_err().map(|val| {
            info!("consumers returned: {:?}", val);
        })
    })
    .map_err(|e| {
        error!("got error: {:?}", e);
        e
    })
    .wait()
    .unwrap();
```

### DIFF
--- a/client/src/connection_manager.rs
+++ b/client/src/connection_manager.rs
@@ -1,0 +1,238 @@
+use crate::connection::{Authentication, Connection};
+use crate::error::ConnectionError;
+use futures::{
+    future::{self, Either},
+    sync::{mpsc, oneshot},
+    Future, Stream,
+};
+use std::collections::HashMap;
+use std::net::SocketAddr;
+use std::sync::Arc;
+use tokio::runtime::TaskExecutor;
+
+/// holds connection information for a broker
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct BrokerAddress {
+    /// IP and port (using the proxy's if applicable)
+    pub address: SocketAddr,
+    /// pulsar URL for the broker
+    pub broker_url: String,
+    /// true if we're connecting through a proxy
+    pub proxy: bool,
+}
+
+/// Look up broker addresses for topics and partitioned topics
+///
+/// The ConnectionManager object provides a single interface to start
+/// interacting with a cluster. It will automatically follow redirects
+/// or use a proxy, and aggregate broker connections
+#[derive(Clone)]
+pub struct ConnectionManager {
+    tx: mpsc::UnboundedSender<Query>,
+    pub address: SocketAddr,
+}
+
+impl ConnectionManager {
+    pub fn new(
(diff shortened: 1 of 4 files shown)


stearnsc:
Sorry for the delay on this; been busy and wanted to make sure I had time to go through and actually understand everything :).

Geal:
Sure, thanks :)

PR #23, opened 2019-04-30
Client api implementation

Geal:
this is the other part for #20: a wrapper API that holds a connection manager and service discovery, can look up topics, create consumers and producers, and even send a single message to a topic in one function call.
Side note: I had an issue with the `create_partitioned_consumers`. It's not possible to share the deserialization closure as it is between all the consumers. So I found it easier to make the `DeserializeMessage` trait and require it on the output type. Maybe that's something that could be integrated into the consumers directly? With a blanket implementation for types that implement `serde::Deserialize`

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -1,0 +1,217 @@
+use crate::connection::Authentication;
+use crate::connection_manager::{BrokerAddress, ConnectionManager};
+use crate::consumer::Consumer;
+use crate::error::{ConsumerError, Error};
+use crate::message::proto::{
+  command_subscribe::SubType, CommandSendReceipt};
+use crate::message::Payload;
+use crate::producer::Producer;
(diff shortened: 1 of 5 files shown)


stearnsc:
>So I found it easier to make the DeserializeMessage trait and require it on the output type. Maybe that's something that could be integrated into the consumers directly? With a blanket implementation for types that implement serde::Deserialize

Without digging in too deeply, that seems fine. Additionally I'd love to move toward making serde a feature rather than core integrated (I had done the `send_json` stuff only because it solved the short-term need I had). I'll open an issue to improve and abstract serialization / deserialization.

stearnsc:
Made https://github.com/wyyerd/pulsar-rs/issues/24 to track

Geal:
Awesome, thanks

PR #27, opened 2019-06-18
support properties on produced messages

Geal:
messages can have a list of properties attached, a map of string to string


client/src/connection.rs
```diff
@@ -529,8 +531,13 @@ pub(crate) mod messages {
         producer_name: String,
         sequence_id: u64,
         num_messages: Option<i32>,
-        data: Vec<u8>
+        data: Vec<u8>,
+        properties: Option<HashMap<String, String>>,
     ) -> Message {
+        let properties = properties.map(|mut h| h.drain().map(|(key, value)| {
```
stearnsc, line 537: What's the benefit of using `drain` here instead of `into_iter()`?

stearnsc (approved):
Looks good, thanks!

PR #32, opened 2019-06-27
improve viz

stearnsc:
Adds an ability to get a stream of `ConsumerState`, which contains metadata (how many topics connected, how many messages received, etc).

### DIFF
--- a/client/src/consumer.rs
+++ b/client/src/consumer.rs
@@ -425,7 +425,7 @@ impl<T: DeserializeMessage> MultiTopicConsumer<T> {
         }
     }
 
-    pub fn get_state_stream(&mut self) -> impl Stream<Item=ConsumerState, Error=()> {
+    pub fn start_state_stream(&mut self) -> impl Stream<Item=ConsumerState, Error=()> {
         let (tx, rx) = unbounded();
         self.state_streams.push(tx);
         rx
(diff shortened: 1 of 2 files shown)


client/src/consumer.rs
```diff
@@ -366,16 +372,27 @@ impl<'a> ConsumerBuilder<'a, Set<Regex>, Set<String>, Set<SubType>> {
     }
 }
 
+/// Details about the current state of the Consumer
+#[derive(Debug, Clone)]
+pub struct ConsumerState {
+    pub connected_topics: Vec<String>,
```
kestred, line 378: Since this struct is cloned for every message received, I'm wondering if it might make sense to use something like an `Arc<[String]>` to avoid unnecessary allocation when creating one.

For a small number of topics, you probably won't notice, but if there were like 100+ topics on the consumer that would probably start to be noticeable.

client/src/consumer.rs
```diff
@@ -402,8 +419,54 @@ impl<T: DeserializeMessage> MultiTopicConsumer<T> {
                 .map_err(|e| panic!("error creating referesh timer: {}", e))),
             subscription: subscription.into(),
             sub_type,
+            last_message_received: None,
+            messages_received: 0,
+            state_streams: vec![],
+        }
+    }
+
+    pub fn get_state_stream(&mut self) -> impl Stream<Item=ConsumerState, Error=()> {
```
kestred: Not important at all, but  its always  weird for me to see a function like `get_XYZ` which takes `&mut self`.
I don't know if another verb could work like `register`, `create`, or `start`.

Feel free to ignore though if none of those sound better to you.

kestred (approved):
🥗

PR #34, opened 2019-07-09
Update README.md to reflect the latest API

ypt:
I ran into a few API changes while exploring pulsar-rs - this PR attempts to update the readme to reflect those.


stearnsc (approved):
I'm hoping to move towards using the `Pulsar` client to create producers and consumers rather than creating them directly, but since I haven't had time to finish those changes, and in the interest of keeping the ball rolling, 👍 (thanks for taking the time to do this)

Issue #40, opened 2019-08-29
Producer and Consumer options


Geal:
there are various options that could be activated on producers and consumers, like priority levels, metadata, schema. I'd like to add a `ProducerOptions` and `ConsumerOptions` arguments at creation, that will match to the optional elements of `CommandSubscribe` and `CommandProducer`. Making separate structures allow easier documentation, and adding more logic when converting to the protobuf structures

PR #41, opened 2019-08-29
Producer and Consumer options

Geal:
I started the work on #40, to see how it would behave in practice. I think a few parameters can be moved in `ConsumerOptions`, like batch size and consumer name

### DIFF
--- a/client/examples/consumer.rs
+++ b/client/examples/consumer.rs
@@ -13,7 +13,7 @@ use futures::{
 use std::net::SocketAddr;
 use tokio::runtime::Runtime;
 
-use pulsar::{message::Payload, Authentication, DeserializeMessage, Pulsar, SubType};
+use pulsar::{message::Payload, Authentication, DeserializeMessage, Pulsar, SubType, ConsumerOptions};
 
 #[derive(Debug, Deserialize)]
 pub struct Data {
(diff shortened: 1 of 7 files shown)


Geal:
@stearnsc this could be merged right now, with this caveat remaining here: https://github.com/wyyerd/pulsar-rs/pull/41/files#diff-3c8b151b706882c736a16732ef666e45R162
the problem comes from the `send_message` implementation, which requests that some producer sends a message on a topic, without specifying which producer: https://github.com/wyyerd/pulsar-rs/blob/master/client/src/producer.rs#L97-L100
I opened #49 to track this problem

stearnsc:
I had to make some editorial changes in resolving merge conflicts, so take a look and let me know if the result still meets your needs.

Some changes: when creating a multi-topic producer, the producer options provided are used for each producer created, rather than the Default impl. From your comment above, I'm assuming you see some use case for setting separate options for each producer (or each topic?)? My vision is roughly that if you want producers to use different options, you have to set it up manually, and not share the same multi-topic producer. Let me know if I'm missing something though; as I said before, I'm not sure I fully grok the use case here.

stearnsc:
If the current state works for you, I'll publish these changes as 0.3 tomorrow

PR #50, opened 2019-11-27
update dependencies

Geal:
this one should not be too hard to review :)
How about a 0.3.0 release with this PR, #41 and #45? Then making the futures 0.3 interface for the next major release?


stearnsc:
Hey, sorry been swamped at work, so a bit behind. Looking at these now, but that seems like a good idea to me.

stearnsc (approved):

PR #56, opened 2020-01-20
Add custom executor support

LucioFranco:

PR #57, opened 2020-01-21
Generic Executor

leshow:
builds on #56 to thread through generic executors

Is this what you were going for @LucioFranco? It required quite a few changes. I removed some unneeded clones & conversions while I was at it.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -12,7 +12,7 @@ use crate::connection::Authentication;
 use crate::connection_manager::{BrokerAddress, ConnectionManager};
 use crate::consumer::{Consumer, ConsumerBuilder, ConsumerOptions, MultiTopicConsumer, Unset};
 use crate::error::Error;
-use crate::executor::PulsarExecutor;
+use crate::executor::{PulsarExecutor, TaskExecutor};
 use crate::message::proto::{self, command_subscribe::SubType, CommandSendReceipt};
 use crate::message::Payload;
 use crate::producer::{self, Producer, ProducerOptions, TopicProducer};
@@ -85,21 +85,20 @@ impl SerializeMessage for str {
 //TODO add more DeserializeMessage impls
 
 #[derive(Clone)]
-pub struct Pulsar<P> {
-    manager: Arc<ConnectionManager<P>>,
+pub struct Pulsar {
+    manager: Arc<ConnectionManager>,
     service_discovery: Arc<ServiceDiscovery>,
-    executor: P,
+    executor: TaskExecutor,
 }
 
-impl<P> Pulsar<P>
-where
-    P: PulsarExecutor,
-{
-    pub fn new(
+impl Pulsar {
+    pub fn new<E: PulsarExecutor>(
         addr: SocketAddr,
         auth: Option<Authentication>,
-        executor: P,
+        executor: E,
     ) -> impl Future<Item = Self, Error = Error> {
+        let executor = TaskExecutor::new(executor);
+
         ConnectionManager::new(addr, auth, executor.clone())
             .from_err()
             .map(|manager| {
@@ -154,7 +153,7 @@ where
             .map(|topics| topics.topics)
     }
 
-    pub fn consumer(&self) -> ConsumerBuilder<Unset, Unset, Unset, P> {
+    pub fn consumer(&self) -> ConsumerBuilder<Unset, Unset, Unset> {
         ConsumerBuilder::new(self)
     }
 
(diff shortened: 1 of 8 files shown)


src/client.rs
```diff
@@ -85,19 +85,22 @@ impl SerializeMessage for str {
 //TODO add more DeserializeMessage impls
 
 #[derive(Clone)]
-pub struct Pulsar {
-    manager: Arc<ConnectionManager>,
+pub struct Pulsar<P> {
```
LucioFranco: very minor nit: I generally like to name the generic closer to the true generic thing, so in this case its an executor so `E` makes much more sense to me.

src/executor.rs
```diff
@@ -0,0 +1,45 @@
+use futures::future::{ExecuteError, Executor, Future};
+use std::sync::Arc;
+
+pub trait PulsarExecutor: Executor<BoxSendFuture> + Send + Sync + 'static + Clone {
```
LucioFranco: So I don't think I originally added clone here because the idea was to quickly convert this trait into `TaskExecutor` then use its arc to be cloneable. So maybe it makes sense to reduce a lot of the passing the generic executor around in favor of quickly changing it to the concrete type of `TaskExecutor`.

LucioFranco (commented):
Overall looks good, left two comments :)

src/executor.rs
leshow: Yeah sorry, I think I just added that to get to the next error and forgot to fix it later. So we want all parts that touch the public API to expose the polymorphic interface but everything else internally can use `TaskExecutor`?

LucioFranco: Yup, thats the idea :)

leshow: Ah! It just clicked what the trick was, two questions,

If `Connection`, for example, holds a `TaskExecutor` 
```rust
pub struct Connection {
    addr: String,
    sender: ConnectionSender,
    sender_shutdown: Option<oneshot::Sender<()>>,
    receiver_shutdown: Option<oneshot::Sender<()>>,
    executor: TaskExecutor,
}
```

We will have a method `new<E: PulsarExecutor>(...)`, but also a method `fn executor(&self) -> TaskExecutor`, right? In that case we may be leaking the internal type. I can change it to `-> impl PulsarExecutor`?

The other thing, calling `Arc::new(executor)`, that allocates a new `Arc` every time we need to get a reference to `TaskExecutor`? If that is in `Connection::new` doesn't that mean every new connection will allocate a new `Arc`?

That was more than two, sorry!

src/client.rs
leshow: This should be resolved now

src/executor.rs
LucioFranco: I think its fine to expose TaskExecutor for now, I think this crate is getting an upgrade so not a big deal.

I think allocating a new Arc per connection is fine for now but if we are getting `TaskExecutor` maybe it makes sense to create a new fn that takes a `TaskExecutor`?

leshow: Don't we want to expose `new` with the `PulsarExecutor` trait so that `vector` is able to implement it? Let me know what you think of the PR as it is now, appreciate your time

LucioFranco: We would need to pass an executor that is generic, it probably should be public but there is blanket impl so nothing can implement that trait but they must implement the proper traits. So vectors executor already implements this trait, ill take a peek at this tomorrow. Thanks!

leshow: Yes, you're right. I forgot about the blanket impl. Let me know if there are any other changes you'd like to see. I'm open to helping `pulsar` update to `std::futures` also if you folks need any help.

LucioFranco: Yeah, I think for now we don't need std::futures support but down the line when we update well prob want that.

LucioFranco (approved):

stearnsc:
Hey, sorry I haven't been plugged into this work. Is this ready for review?

leshow:
I believe so, I've tested this locally while building the pulsar sink for vector and things are compiled & exposed correctly. I could quickly drop in #60 w/ the `?Sized` bound also if you like.

src/client.rs
```diff
@@ -268,7 +270,7 @@ impl Pulsar {
             })
     }
 
-    pub fn create_producer<S: Into<String>>(
+    pub fn create_producer<S: Into<String>, E: PulsarExecutor>(
```
stearnsc: I'm not sure I'm understanding why the methods that don't themselves take a `E: PulsarExecutor` need the `E` type param. Can you clarify? I'm referring to `Pulsar::create_producer`, `Pulsar::create_partitioned_producers`, `TopicProducer::from_connection`, `Producer::send`, and `Producer::send_raw`

leshow: I think you're right, I may have confused them with `ConnectionManager::from_connection`. Let me try to get rid of them and update the PR.

leshow: I've rebased from master and fixed these, you're right they are unnecessary.

stearnsc (approved):

PR #58, opened 2020-01-21
Format using 'cargo fmt'

Atul9:


stearnsc:
fixed in https://github.com/wyyerd/pulsar-rs/pull/66

PR #63, opened 2020-02-06
Add support for batched messages

### LINKED ISSUE #59: Handling batched messages
The current implementation does not deal with batched messages. When pulsar sends a batch of messages, the message payload will contain a sequence of `SingleMessageMetadata` followed by the individual message payload, for each message.

Due to this, a naive implementation of  `DeserializeMessage` will fail when receiving a batched message, as it has to read the `SingleMessageMetadata` before processing the payload.

This also requires #36 to be resolved, as the entire batch must be decompressed.

gperinazzo:
Closes #59 

Adds support for batched messages:
- Batched messages are properly separated and exposed as single messages
- `AckHandler` now tracks batched messages to ack/nack only once when either all messages are acked or one is nacked

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ log = "0.4.6"
 trust-dns-resolver = "0.12"
 url = "2.1"
 regex = "1.1.7"
+bit-vec = "0.6"
 
 [dev-dependencies]
 serde = { version = "1.0.101", features = ["derive"] }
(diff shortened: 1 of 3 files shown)


Geal:
Wow that's great, I could really use message batching

NotBad4U:
nice !

stearnsc (approved):

PR #71, opened 2020-04-16
Futures 0.3

Geal:
it is not finished yet, but at least it compiles and the tests run :smile: 
It includes dependency updates for bytes, prost, prost-derive, tokio, futures-timer and trust-dns-resolver since those were also converted to the new futures.
I converted a large part of the code to async/await since it solves some compilation issues, an overall it makes the code a bit easier to follow


Geal:
ok, this should be usable now :)

Geal:
I'm merging this now, after more tests things are looking good, I'll have to investigate further design issues

PR #75, opened 2020-04-23
Format code using 'cargo fmt'

Atul9:


Geal:
please do not do that, it does not help much. If you have time to send those kind of pull requests, I'd prefer you write examples and documentation

PR #77, opened 2020-05-03
Expose TokioExecutor

vmalloc:
Currently creating a Pulsar instance within a tokio app is impossible given that TokioExecutor is private. This PR exposes it so that it can be used


Geal:
thanks!

PR #84, opened 2020-05-11
implement compression

Geal:
this commit chain will add support for compression and decompression:
- [x] LZ4
- [x] zlib
- [x] zstd
- [x] snappy


Geal:
all the compression algorithms are now integrated. That wasn't too hard :sweat_smile:

PR #88, opened 2020-05-20
async-std support

Geal:
first commit: remove the executor member, and instead get the executor statically (to do  `tokio::runtime::Handle::current().spawn(f)` with tokio and `task::spawn(f)` wih async-std).

@LucioFranco IIRC you did the initial work to make pulsar-rs generic over the executor (was that to support tokio 0.1 and tokio 0.2 in vector?). Will this change cause issues with the integration in vector?

From a user's point of view, the main change is this https://github.com/wyyerd/pulsar-rs/pull/88/files#diff-f5be2bf3fd69683200ad4f58e2550984R63

(and I'll add `new_tokio` an `new_async_std` methods to simplify that)

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -1,15 +1,29 @@
 use futures::future::{Future, FutureExt};
-use std::sync::Arc;
-use std::pin::Pin;
 use std::ops::Deref;
+use std::pin::Pin;
+use std::sync::Arc;
 use tokio::runtime::Handle;
 
 pub trait Executor: Send + Sync {
     fn spawn(&self, f: Pin<Box<dyn Future<Output = ()> + Send>>) -> Result<(), ()>;
 }
 
+pub enum ExecutorKind {
+    Tokio,
+    AsyncStd,
+}
+
 pub trait PulsarExecutor: Clone + Send + Sync + 'static {
     fn spawn(f: Pin<Box<dyn Future<Output = ()> + Send>>) -> Result<(), ()>;
+    fn spawn_blocking<F, Res>(f: F) -> JoinHandle<Res>
+    where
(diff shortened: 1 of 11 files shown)


src/executor.rs
```diff
@@ -8,6 +8,10 @@ pub trait Executor: Send + Sync {
     fn spawn(&self, f: Pin<Box<dyn Future<Output = ()> + Send>>) -> Result<(), ()>;
 }
 
+pub trait PulsarExecutor: Clone + Send + Sync + 'static {
```
LucioFranco: you could probably add `?Sized` here too

LucioFranco (commented):
Yes, we have a custom runtime that isn't available right away but this should work.

Geal:
is it a runtime that works on tokio 0.1 or 0.2? I have another issue here with TCP connections: since tokio and async-std each come with their version of `TcpStream`, I'd need to abstract that.
Unfortunately, it would be a trait method that would be async (or at least return Impl Future or Box<Pin<Future<etc>>>), and async trait methods are not available right now.
Here are the solutions I have:
- use the trait-async crate to work around the trait method
- have the trait provide a `kind()` method that returns a `enum ExecutorKind { Tokio, AsyncStd }` and decide at runtime which version will be used

I'm currently working on the second option, it's a bit ugly but the code is straightforward and supporting the two main runtimes should be enough for now, I think. Any opinions on this?

Geal:
alright, it's in good shape, the only thing remaining now is abstracting `interval()` and `timeout()`

Geal:
ok, it's working now! Next I'll look into making tokio or async-std usage optional through a cargo feature

PR #89, opened 2020-05-21
test for github actions CI

Geal:


Geal:
so apparently it works, but there's a weird behaviour: the tests never stop :thinking:

Geal:
alright, it works now. The issue was something I have seen a few times already: if the Pulsar instance was just created, and I create the consumer after the producer, the consumer could hang waiting for the message. I do not know yet where this is coming from

Geal:
that was caused by consumers created with "latest" for their initial position by default, so they only get new messages sent on the topic after the consumer's creation

PR #95, opened 2020-05-25
Reconnection

Geal:
this PR starts the work to handle reconnections while keeping existing Producer and Consumer instances. Part of the difficulty comes from sharing the `Connection`: replacing the connection with a new one means mutable access to the fields of `ConnectionSender`. I could put a Mutex around `ConnectionSender` or some of its fields, but there's a risk of deadlock if I'm not careful with the async code. Another option would be to hide mutability in yet another task


Geal:
I am now seeing the plan more clearly, and it's doable, but requires a bit of work:

For every message we can send on a connection (basically, the methods of `ConnectionSender`), we need to wrap it that way: https://github.com/wyyerd/pulsar-rs/pull/95/commits/6d0ad4018591152ad10b9b73d44fed38cde4554c#diff-82315f246b58c7df91237127e92e65e5R42-R50

For producers and consumers, it is more complex, since we will need to recreate them after the new connection is established, then send the message again.

For producers, we ill need to store all the messages for which we have not received a receipt yet, resend them all **in order**, then send the message again.
For consumers, we will need to ask for unacked messages to be redelivered. I'm not sure the messages will arrive in the proper order

Geal:
alright it's starting to work properly. The remaining steps:
- [x] making the exponential backoff variables configurable
- [ ] handling some edge cases like:
  - [x] getting a `Disconnected` error when calling `send_flow` while handling messages from a subscription (we must reconnect and retry)
  - [ ] getting a `Disconnected` error when sending acks (reconnect, retry, and send back an error if we still failed)
  - [ ] keeping messages in a queue in the producer until we get the receipts for them (so we can resend them if we reconnect)
  - [ ] doing backoff per IP (right now it's over any DNS result)
- [x] implementing reconnection for `Pulsar` methods (that part is much easier than producers and consumers)
- [x] reactivating and testing negative acknowedgements

There will probably be bugs since a reconnection is a complex event, but the current behaviour is already very stable. The "round trip" example can recover its producer and consumer even while waiting 10s for a standalone pulsar to restart. In more common cases, there will be multiple brokers, and since we do a DNS lookup on each connection creation we can end up on a different server rather quickly.

There are a few API changes required here:
- `Consumer::ack()` is now async
- `Producer::send()` requires the producer to be mutable (it must be able to modify its own connection member

PR #109, opened 2020-07-04
Owned serialize

Geal:
trying out a solution to avoid copies when serializing (cf https://github.com/wyyerd/pulsar-rs/issues/67#issuecomment-653703530 )

### DIFF
--- a/examples/batching.rs
+++ b/examples/batching.rs
@@ -12,8 +12,8 @@ struct TestData {
 }
 
 impl SerializeMessage for TestData {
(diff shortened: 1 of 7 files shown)


Geal:
It also adds an impl of `SerializeMessage` for `producer::Message` so we can write:

```rust
producer.send(producer::Message {
  payload: Vec::new(),
  ..Default::default()
}).await
```

stearnsc:
I like this better. Makes it easy to send a custom message, and easy to send a byte array without copying

Issue #107, opened 2020-07-02
Dead Letter Queue policy for Consumer


rtakasu:
Is there any plan to implement a DLQ policy for the Consumer? Similar to what other client libraries have (eg. DeadLetterPolicy for Java client: https://pulsar.apache.org/docs/en/client-libraries-java/)

Geal:
yes, but probably not in this release (we need to ship that 1.0 soon)

stearnsc:
Closed by https://github.com/wyyerd/pulsar-rs/pull/110

PR #110, opened 2020-07-07
Feature/dlq

rtakasu:


rtakasu:
Took a stab at Dead Letter Queues. Related to #107

src/consumer.rs
```diff
@@ -573,35 +583,75 @@ impl<Exe: Executor + ?Sized> ConsumerEngine<Exe> {
...
         }
 
         Ok(())
     }
 
+    async fn send_to_consumer(
+        &mut self,
+        message_id: MessageIdData,
+        payload: Payload,
+        request_time: Instant,
```
stearnsc: This would probably be easier to read initialized in the function rather than as an arg, since it's always `Instant::now()` (and would be confusing to be anything else)

src/consumer.rs
```diff
@@ -1550,4 +1630,95 @@ mod tests {
 
         rt.block_on(f);
     }
+
+    #[test]
```
stearnsc: In https://github.com/wyyerd/pulsar-rs/pull/112 I went through and cleaned up these tests to use `#[tokio::test]` - I think it makes them easier to read / maintain. Might be worth taking a look and doing that here as well.

src/consumer.rs
```diff
@@ -1550,4 +1630,95 @@ mod tests {
...
+    #[cfg(feature = "tokio-runtime")]
+    fn dead_letter_queue() {
+        use rand::{distributions::Alphanumeric, Rng};
+        let _ = log::set_logger(&TEST_LOGGER);
+        let _ = log::set_max_level(LevelFilter::Debug);
+        let addr = "pulsar://127.0.0.1:6650";
+        let mut rt = Runtime::new().unwrap();
+
+        let topic = "dead_letter_queue_test";
```
stearnsc: I've started making these topics include a random u16 in order to remove stale data from previous runs of the test. It might be worth doing that here, something like
```
let test_id: u16 = rand::random();
let topic = format!("dead_letter_queue_test_{}, test_id);
...
let dead_letter_topic = format!("{}_dlq", &topic);
```

stearnsc (approved):
This looks great! Super happy to have this in before the 1.0 release. I've made some small suggestions, but nothing API breaking so I think this can be merged as-is.

src/consumer.rs
rtakasu: I agree. The only reason why I left the initialization outside the function was because of the batched message use case, where it was previously doing `now = Instant::now()` and then using the same value for all iterations of the loop. [link](https://github.com/wyyerd/pulsar-rs/blob/master/src/consumer.rs#L574)

Do you think its ok for the batched message to have slightly different values in the `unacked_messages` map?

stearnsc: Hm, I think ideally the whole batch would go off of the timestamp of the _last_ message, but there's no way to do that. Probably having them have different timestamps then makes the most sense, because it at least prevents the case of a large batch and a slow consumer causing the timeout to be reached before the last message has been received.

rtakasu:
Addressed the comments and made some updates to the test to prevent it from picking up old runs ✅

stearnsc:
Closes https://github.com/wyyerd/pulsar-rs/issues/107

Geal:
that's great, thanks!

PR #112, opened 2020-07-07
Cleanup API

stearnsc:
Major changes: 
1. Everything is only constructible from a builder. This allows us to add improvements without breaking the API
2. The phantom types on the builders have been removed
3. Both consumers and producers handle partitions transparently
4. The same public consumer type is used for regex-matched consumers vs explicit topic consumers
5. The client's `send` is backed by the MultiTopicProducer
6. Tests cleaned up and simplified
7. All doc-tests compile
8. (most) implementation details are no longer public (`messages` still are).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pulsar"
-version = "1.0.0"
+version = "1.0.0-alpha-2"
 edition = "2018"
 authors = [
     "Colin Stearns <cstearns@developers.wyyerd.com>",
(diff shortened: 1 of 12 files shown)


stearnsc:
@Geal This contains some major surgery and some editorial choices. Some of the implementation is the "lazy" solution (using an enum to back different producer/consumer types rather than doing anything more clever), but we can always clean that up later. I particularly would love your feedback on the API and documentation.

Geal:
the API looks much better, especially the builders. And now that the public API focuses on a few builders the documentation is much easier to follow, everything is ok for me. Also using a multi topic producer in `Pulsar::send` was a good idea, I suspect this function will be used a lot

PR #113, opened 2020-07-09
executor: Store an instance instead of using raw methods

Keruspe:
This allows some consumer of pulsar-rs to define a custom executor that can carry some state around

It also makes things easier when some software (vector being an exemple) already carries an object around to abstract the executor on top of some other one. This way we can just pass that object around.

### DIFF
--- a/examples/batching.rs
+++ b/examples/batching.rs
@@ -34,7 +34,7 @@ async fn main() -> Result<(), pulsar::Error> {
     env_logger::init();
 
     let addr = "pulsar://127.0.0.1:6650";
-    let pulsar: Pulsar<TokioExecutor> = Pulsar::builder(addr).build().await?;
+    let pulsar: Pulsar<_> = Pulsar::builder(addr, TokioExecutor).build().await?;
     let mut producer = pulsar
         .producer()
         .with_topic("test-batch-compression-snappy")
--- a/examples/consumer.rs
+++ b/examples/consumer.rs
@@ -24,7 +24,7 @@ async fn main() -> Result<(), pulsar::Error> {
     env_logger::init();
 
     let addr = "pulsar://127.0.0.1:6650";
-    let pulsar: Pulsar<TokioExecutor> = Pulsar::builder(addr).build().await?;
+    let pulsar: Pulsar<_> = Pulsar::builder(addr, TokioExecutor).build().await?;
 
     let mut consumer: Consumer<TestData, _> = pulsar
         .consumer()
--- a/examples/producer.rs
+++ b/examples/producer.rs
@@ -24,7 +24,7 @@ async fn main() -> Result<(), pulsar::Error> {
     env_logger::init();
 
     let addr = "pulsar://127.0.0.1:6650";
-    let pulsar: Pulsar<TokioExecutor> = Pulsar::builder(addr).build().await?;
+    let pulsar: Pulsar<_> = Pulsar::builder(addr, TokioExecutor).build().await?;
     let mut producer = pulsar
         .producer()
         .with_topic("non-persistent://public/default/test")
--- a/examples/round_trip.rs
+++ b/examples/round_trip.rs
@@ -34,7 +34,7 @@ async fn main() -> Result<(), pulsar::Error> {
     env_logger::init();
 
     let addr = "pulsar://127.0.0.1:6650";
-    let pulsar: Pulsar<TokioExecutor> = Pulsar::builder(addr).build().await?;
+    let pulsar: Pulsar<_> = Pulsar::builder(addr, TokioExecutor).build().await?;
     let mut producer = pulsar
         .producer()
         .with_topic("test")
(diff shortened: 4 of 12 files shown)


Geal:
for some context: the library previously stored the executor and I changed to a trait in 097866ea1d85aef55c2574511564bc94a5e4280b, because I thought I could isolate the few parts that use the executor, but after implementing reconnections, it appears the executor must be everywhere.
This change will ease future integrations with other executors that would wrap tokio or async-std, like https://github.com/timberio/vector is doing.
@stearnsc I know it is coming a bit late while we're trying to release a 1.0, but it should not break existing functionality, it would only change the API

Geal:
I just tried with a small wrapper for the tokio executor, to count the number of spawned tasks (and while I'm at it, try the new `track_caller` feature) in the "round_trip" example:

```rust
#![feature(track_caller)]
#[macro_use]
extern crate serde;
use futures::TryStreamExt;
use pulsar::{
    message::proto, message::proto::command_subscribe::SubType, message::Payload, producer,
    Consumer, DeserializeMessage, Error as PulsarError, Pulsar, SerializeMessage,
    executor::{JoinHandle, Delay, Interval, ExecutorKind}, Executor,
};
use std::{sync::{Arc, Mutex}, pin::Pin, future::Future};
use std::panic::Location;

/// Wrapper for the Tokio executor
#[cfg(feature = "tokio-runtime")]
#[derive(Clone, Debug)]
pub struct MyExecutor(pub Arc<Mutex<usize>>);

#[cfg(feature = "tokio-runtime")]
impl Executor for MyExecutor {
    #[track_caller]
    fn spawn(&self, f: Pin<Box<dyn Future<Output = ()> + Send>>) -> Result<(), ()> {
        {
            let mut counter = self.0.lock().unwrap();
            *counter += 1;
            println!("task n°{} spawned from {}", *counter, Location::caller());
        }
        tokio::task::spawn(f);
        Ok(())
    }

    #[track_caller]
    fn spawn_blocking<F, Res>(&self, f: F) -> JoinHandle<Res>
    where
        F: FnOnce() -> Res + Send + 'static,
        Res: Send + 'static,
    {
        {
            let mut counter = self.0.lock().unwrap();
            *counter += 1;
            println!("blocking task n°{} spawned from {}", *counter, Location::caller());
        }
        JoinHandle::Tokio(tokio::task::spawn_blocking(f))
    }

    fn interval(&self, duration: std::time::Duration) -> Interval {
        Interval::Tokio(tokio::time::interval(duration))
    }

    fn delay(&self, duration: std::time::Duration) -> Delay {
        Delay::Tokio(tokio::time::delay_for(duration))
    }

    fn kind(&self) -> ExecutorKind {
        ExecutorKind::Tokio
    }
}

#[derive(Serialize, Deserialize)]
struct TestData {
    data: String,
}

impl SerializeMessage for TestData {
    fn serialize_message(input: Self) -> Result<producer::Message, PulsarError> {
        let payload = serde_json::to_vec(&input).map_err(|e| PulsarError::Custom(e.to_string()))?;
        Ok(producer::Message {
            payload,
            ..Default::default()
        })
    }
}

impl DeserializeMessage for TestData {
    type Output = Result<TestData, serde_json::Error>;

    fn deserialize_message(payload: &Payload) -> Self::Output {
        serde_json::from_slice(&payload.data)
    }
}

#[tokio::main]
    env_logger::init();
    let executor = MyExecutor(Arc::new(Mutex::new(0)));

    let addr = "pulsar://127.0.0.1:6650";
    let pulsar: Pulsar<_> = Pulsar::builder(addr, executor.clone()).build().await?;
    let mut producer = pulsar
        .producer()
        .with_topic("test")
        .with_name("my-producer")
        .with_options(producer::ProducerOptions {
            schema: Some(proto::Schema {
                type_: proto::schema::Type::String as i32,
                ..Default::default()
            }),
            ..Default::default()
        })
        .build()
        .await?;

    tokio::task::spawn(async move {
        let mut counter = 0usize;
        loop {
            producer
                .send(TestData {
                    data: "data".to_string(),
                })
                .await
                .unwrap()
                .await
                .unwrap();
            counter += 1;
            if counter % 1000 == 0 {
                println!("sent {} messages", counter);
            }
        }
    });

    let pulsar2: Pulsar<_> = Pulsar::builder(addr, executor).build().await?;

    let mut consumer: Consumer<TestData, _> = pulsar2
        .consumer()
        .with_topic("test")
        .with_consumer_name("test_consumer")
        .with_subscription_type(SubType::Exclusive)
        .with_subscription("test_subscription")
        .build()
        .await?;

    let mut counter = 0usize;
    while let Some(msg) = consumer.try_next().await? {
        consumer.ack(&msg).await?;
        let data = msg.deserialize().unwrap();
        if data.data.as_str() != "data" {
            panic!("Unexpected payload: {}", &data.data);
        }
        counter += 1;
        if counter % 1000 == 0 {
            println!("received {} messages", counter);
        }
    }

    Ok(())
}
```
It's easy to use and does not change the API much. The only thing missing is exposing the executor module (so one can use the types when implementing the `Executor` trait)

stearnsc:
I haven't worked on any of the executor code so far, but right now it's not possible to impl a custom executor because of the `ExecutorKind` requirement, right? Have we explored using `async-trait` and getting rid of that enum altogether?

Geal:
`async-trait` puts every call in a box, last time I tried it could easily become costly (especially considering that the executor has to box things up too). I think we'll have to wait a bit for solutions to put tokio or async-std behind the same interface (iirc people are working on it)

PR #120, opened 2020-11-03
Remove ack_tx from ConsumerEngine

### LINKED ISSUE #119: Consumer is never dropped
It looks like consumers are never closed after being created. This becomes an issue if you drop an exclusive consumer and then try to re-open it. Here's an example that reproduces the issue:
```rust
use pulsar::{Consumer, Pulsar, SubType, TokioExecutor};
use std::time::Duration;
use tokio::time::delay_for;

#[tokio::main]
    env_logger::init();

    let addr = "pulsar://127.0.0.1:6650";
    let pulsar: Pulsar<_> = Pulsar::builder(addr, TokioExecutor).build().await?;

    loop {
        println!("Attempting to create consumer");
        let result: Result<Consumer<Vec<u8>, _>, pulsar::Error> = pulsar
            .consumer()
            .with_topic("non-persistent://public/default/test")
            .with_consumer_name("test_consumer")
            .with_subscription_type(SubType::Exclusive)
            .with_subscription("test_subscription")
            .build()
            .await;

        let consumer = match result {
            Ok(c) => c,
            Err(e) => {
                eprintln!("Failed to create consumer: {}", e);
                delay_for(Duration::from_secs(1)).await;
                continue;
            }
        };

        delay_for(Duration::from_secs(1)).await;
        drop(consumer);
        delay_for(Duration::from_secs(1)).await;
    }
}
```
The first consumer will be created, but it's never dropped, and every following attempt fails with `Connection error: Server error (Some(ConsumerBusy)): Exclusive consumer is already connected`

gperinazzo:
The `ConsumerEngine` uses the ack receiver to know when it should stop (Consumer object was dropped). Passing the sender into the engine introduced a bug where dropping the consumer object no longer stops the engine.

This PR changes the place inside the engine where the sender was used to eagerly ack the message instead. Can anyone comment on if this would cause any issues?

Fixes #119

### DIFF
--- a/src/consumer.rs
+++ b/src/consumer.rs
@@ -335,7 +335,6 @@ impl<T: DeserializeMessage> TopicConsumer<T> {
             name,
             tx,
             messages,
-            ack_tx.clone(),
             ack_rx,
             batch_size,
             unacked_message_redelivery_delay,
@@ -446,7 +445,6 @@ struct ConsumerEngine<Exe: Executor> {
     name: Option<String>,
     tx: mpsc::Sender<Result<(proto::MessageIdData, Payload), Error>>,
     messages_rx: Option<mpsc::UnboundedReceiver<RawMessage>>,
-    ack_tx: mpsc::UnboundedSender<AckMessage>,
     ack_rx: Option<mpsc::UnboundedReceiver<AckMessage>>,
     batch_size: u32,
     remaining_messages: u32,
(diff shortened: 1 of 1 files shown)


Geal:
I'll do some tests tomorrow, but this seems fine, thank you for catching that bug

Geal:
Thanks!

PR #122, opened 2020-11-12
Clippy fixes

vmalloc:


Geal:
will this fail the CI if clippy fails?

vmalloc:
Supposed to. Yes

vmalloc:
@Geal I rebased on top of master to fix the latest conflicts

Geal:
thanks. I'd rather have the clippy check non blocking in CI, it can be very noisy.

vmalloc:
Ok, although from my personal experience (especially when pinning the CI to a specific rust version and working with PRs) it becomes negligible. Also, if it doesn't fail the CI the code issues will continue piling up until they're "someone else's problem" trying to just make sure his or her code does not introduce more issues, but that's just my 2 cents

Geal:
I'm not against a linter pass in the CI, but it should be done once we know which linters are actually useful for the project and which are just noise. If it fails all the time for linters we can ignore, the CI will be entirely ignored too

vmalloc:
you mean which lints, not linters right? So the problem is creating a config file with lints to ignore? that sounds like a reasonable thing to have.

About clippy in general I think it tends to be very useful, including lints pointing to actual bugs - and I'm not familiar with an alternative linter for Rust that is considered better or less noisy

Geal:
yes, that's what I mean :)

vmalloc:
@Geal I can continue working on cleaning this up and adding excludes, but I'd hate to waste my time if this ends up not being merged. Assuming I add sane defaults to clippy and fix the remaining issues (that in the mean time continue to pile up due to merged PRs and clippy advancements) - is the general disposition to merge this in?

Geal:
Yes, it will be merged, no worries. It would have been smoother as two PRs: one to fix issues reported by clippy, one to add the CI task. But it will be alright

vmalloc:
Ok. I will update this PR soon with only the clippy fixes. Thanks!

vmalloc:
PR Updated

vmalloc:
@Geal When you get the chance, please review. Thanks!

Geal:
Thanks!

PR #125, opened 2020-11-30
Implemented seek method

DonghunLouisLee:
code is not organized and error handling was not done because adding seek method to the current logic requires recursive error handling..
I've made a simple test for the method and it works as expected
Would love feedbacks


Geal:
Hello! Thank you for the PR, I merged it as 56bd66613384588035958f20d3ad8254e7084892 with the following modifications:
- squashed in one commit
- rebased over the main branch to avoid formatting changes (could you set up your editor so that it does not format automatically everything? It should follow what projects do, and if they don't specify anything, it shouldn't format without asking)
- replaced a few `unwrap()` calls with `?`
- removed the `drop` calls (they were not doing anything since they're called on a reference, and it is dropped correctly anyway when replacing the old consumer)

The `ConsumerEngine` already has a method for reconnection, but it would be a bit cumbersome to use it here, so the reconnection in the PR is fine for now

PR #134, opened 2021-02-05
connection: try all IPs addresses when connecting, instead of a random one

Keruspe:


Geal:
I looked a bit more at this and I'm not convinced it will solve the connections issues we've seen: in my tests the list of IP addresses is always returned in the same order, so the first server would get all of the traffic from all clients, until it cannot accept new traffic and then we go onto the second one. Randomizing is still balancing the traffic better

Keruspe:
Rebased + shuffled the order

FlorentinDUBOIS (approved):

tisonkun:
@FlorentinDUBOIS given it's quite a long time since the patch was made, I'm going to rebase this patch while keeping @Keruspe as the coauthor.

Issue #135, opened 2021-02-11
add a configurable connection timeout


Geal:
we already have retry and backoff policies, but they do not cover the case of slow servers

Geal:
there is now a connection timeout

PR #136, opened 2021-02-12
WIP: Connection timeout

Geal:
implementation for #135. We're currently testing it


Geal:
merged as part of #138

Issue #131, opened 2021-01-28
add a method for setting initial position for ConsumerOptions explicitly


DonghunLouisLee:
Currently, to set the subscription to earliest, you have to do 

.with_options(ConsumerOptions { initial_position: Some(1), ..Default::default() }) 

which is good enough but i think it might be better if we can add a method for changing the initial position explicitly like 
http://pulsar.apache.org/api/client/2.4.2/org/apache/pulsar/client/api/SubscriptionInitialPosition.html

I can work on this if necessary! 
Thanks

Geal:
sure, it would be nice!

Geal:
I added an enum in https://github.com/wyyerd/pulsar-rs/commit/b06e67de5df21eb72b96d98cc0286d408fe1c057
I'll replace the `Option<i32>` in  `ConsumerOptions` on the next major release as it would be a breaking change

PR #137, opened 2021-02-12
inital_position separated

DonghunLouisLee:
implemented #131

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -336,6 +336,7 @@ impl ConnectionSender {
         sub_type: SubType,
         consumer_id: u64,
         consumer_name: Option<String>,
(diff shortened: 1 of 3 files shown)


Geal:
I don't understand, why are you replacing the setting in `ConsumerOptions` everywhere with a new one, instead of changing it only in the builder?

Here's how it could work better:
- keep only the `with_initial_position` method, make it create a `ConsumerOptions` or update the current one with the position
- use as argument the enum you added but commented out, be cause earliest/latest is more explicit than a number

Geal:
implemented in https://github.com/wyyerd/pulsar-rs/commit/b06e67de5df21eb72b96d98cc0286d408fe1c057

Issue #144, opened 2021-03-09
Reader interface


Geal:
Other Pulsar clients have a [Reader](https://pulsar.apache.org/api/client/2.7.0-SNAPSHOT/org/apache/pulsar/client/api/Reader.html), we should add it.

Essentially, it's a wrapper around a consumer with non durable subscription and automatic acks

Geal:
done by @Keksoj in #159

Issue #126, opened 2020-12-28
Cooperation on Pulsar client for Rust


Anonymitaet:
Hi @wyyerd and Stearns,
 
Hope you’re having a great holiday.
 
My name is Yu Liu. I'm a Pulsar committer and technical writer.
 
When browsing the Pulsar website, I found that you have developed a wonderful [Rust client for Pulsar](https://github.com/wyyerd/pulsar-rs) and draw much attention. Thanks for this great work!

Since good projects are collaborations between motivated individuals, **I'm wondering if you'd like to donate the Rust client to Apache Pulsar**, which is beneficial for attracting more programmers to share quality code, appealing to more technical writers to develop docs (I found you've written some guides and I'd like to contribute together), and allowing more users to use it conveniently.
 
Any thoughts on this? Looking forward to hearing from you.
 
Merry Christmas with lots of love.
 
Yu

Geal:
Hello @Anonymitaet,
it would make sense to integrate the Rust client in the Pulsar project and have it used as the official client. So I'm ok with that! How about you @stearnsc ?

What would be the next steps for integration under the Pulsar project?

stearnsc:
Yeah, I'm also in support of this. Can you give us a brief rundown of how this would work @Anonymitaet?

Anonymitaet:
Thank you @Geal and @stearnsc. 

The first step is to write a PIP for this. You can check [PIP 53: Contribute DotPulsar to Apache Pulsar](https://github.com/apache/pulsar/wiki/PIP-53%3A-Contribute-DotPulsar-to-Apache-Pulsar)
as an example and send it or anything you want to know (for example, where the code lives, who maintains it, who publishes it,  and so on) to `dev@pulsar.apache.org`. PMC will handle and help with the next procedures.

Anonymitaet:
Any progress on this? @Geal @stearnsc

stearnsc:
It's a super busy time for us at work right now, but I should get to this in the next few days, most likely.

Anonymitaet:
@stearnsc OK, looking forward to your feedback on PIP.

Anonymitaet:
Hi @stearnsc, 

Since you are working on PIP, maybe we can start working on the user guide as well? I'd like to contribute together and I'm wondering which way do you prefer:
 
[...]

PR #159, opened 2021-05-03
Implement a Reader

Keksoj:
The Apache Software Foundation considers integrating pulsar-rs, see issue #126 

In order to expand the rust client's features, we need to implement a Reader interface, as suggested at issue #144 . A Reader is a pulsar client that immediately aknowledges messages upon reception.

### DIFF
--- a/examples/reader.rs
+++ b/examples/reader.rs
@@ -1,0 +1,87 @@
+#[macro_use]
+extern crate serde;
+use futures::TryStreamExt;
+use pulsar::{
+    consumer::ConsumerOptions, proto::Schema, reader::Reader, Authentication, DeserializeMessage,
+    Payload, Pulsar, TokioExecutor,
+};
+use std::env;
+
+#[derive(Serialize, Deserialize)]
+struct TestData {
+    data: String,
+}
+
+impl DeserializeMessage for TestData {
+    type Output = Result<TestData, serde_json::Error>;
+
+    fn deserialize_message(payload: &Payload) -> Self::Output {
+        serde_json::from_slice(&payload.data)
+    }
+}
+#[tokio::main]
+async fn main() -> Result<(), pulsar::Error> {
+    env_logger::init();
+
+    let addr = env::var("PULSAR_ADDRESS")
+        .ok()
+        .unwrap_or_else(|| "pulsar://127.0.0.1:6650".to_string());
+    let topic = env::var("PULSAR_TOPIC")
+        .ok()
+        .unwrap_or_else(|| "non-persistent://public/default/test".to_string());
+
+    let mut builder = Pulsar::builder(addr, TokioExecutor);
+
+    if let Ok(token) = env::var("PULSAR_TOKEN") {
+        let authentication = Authentication {
+            name: "token".to_string(),
+            data: token.into_bytes(),
+        };
+
+        builder = builder.with_auth(authentication);
+    }
+
+    let pulsar: Pulsar<_> = builder.build().await?;
+
+    let mut reader: Reader<TestData, _> = pulsar
+        .reader()
+        .with_topic(topic)
+        .with_consumer_name("test_reader")
+        .with_options(ConsumerOptions::default().with_schema(Schema {
+            r#type: pulsar::proto::schema::Type::String as i32,
+            ..Default::default()
+        }))
+        // subscription defaults to SubType::Exclusive
+        .into_reader()
+        .await?;
+    // log::info!("created a reader");
+
+    let mut counter = 0usize;
+
+    // listen to 5 messages
+    while let Some(msg) = reader.try_next().await? {
+        log::info!("metadata: {:#?}", msg.metadata());
+
+        log::info!("id: {:?}", msg.message_id());
+        let data = match msg.deserialize() {
+            Ok(data) => data,
+            Err(e) => {
+                log::error!("Could not deserialize message: {:?}", e);
+                break;
+            }
+        };
+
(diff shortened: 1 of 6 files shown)


Geal (changes requested):

Geal (approved):

.gitignore
```diff
@@ -2,3 +2,4 @@
 .idea
 **/*.rs.bk
 Cargo.lock
+.vscode/
```
Geal: this should be removed (we won't start adding special case for every editor)

README.md
```diff
@@ -1,6 +1,8 @@
 ## Pulsar
 #### Future-based Rust client for [Apache Pulsar](https://pulsar.apache.org/)
 
+![under construction](http://www.chipenergy.com/images/Under_Construction.gif)
```
Geal: this should be removed

examples/reader.rs
```diff
@@ -0,0 +1,95 @@
...
+        builder = builder.with_auth(authentication);
+    }
+
+    let pulsar: Pulsar<_> = builder.build().await?;
+
+    let mut reader: Reader<TestData, _> = pulsar
+        .reader()
+        .with_topic(topic)
+        .with_consumer_name("test_reader")
+        .with_options(
```
Geal: why is the reader example setting the durable option and the initial position? Those should be set already when creating the reader(durable to false, initial position to latest)

src/client.rs
```diff
@@ -271,6 +272,15 @@ impl<Exe: Executor> Pulsar<Exe> {
         ProducerBuilder::new(self)
     }
 
+    /// creates a reader builder
+    /// ```rust, no_run
+    /// // please fill out an example
```
Geal: please fill out an example :wink:

src/client.rs
```diff
@@ -271,6 +272,15 @@ impl<Exe: Executor> Pulsar<Exe> {
...
+    /// ```
+    pub fn reader(&self) -> ConsumerBuilder<Exe> {
+        // this makes it exactly the same like the consumer() method though
+        ConsumerBuilder::new(self)
```
Geal: this is where the durable option and the initial position should be set

src/consumer.rs
```diff
@@ -61,6 +62,44 @@ pub struct ConsumerOptions {
     pub initial_position: InitialPosition,
 }
 
+impl ConsumerOptions {
```
Geal, line 65: those methods could be proposed in a separate pull request

src/consumer.rs
```diff
@@ -61,6 +62,44 @@ pub struct ConsumerOptions {
...
+    /// within options, sets the priority level
+    pub fn with_priority_level(mut self, priority_level: i32) -> Self {
+        self.priority_level = Some(priority_level);
+        self
+    }
+
+    pub fn with_durable_option(mut self, durable: bool) -> Self {
```
Geal: we don't need the `_option` suffix

src/consumer.rs
```diff
@@ -1454,10 +1501,9 @@ impl<Exe: Executor> ConsumerBuilder<Exe> {
             consumer_name,
             batch_size,
             unacked_message_resend_delay,
-            namespace,
-            topic_refresh,
             consumer_options,
             dead_letter_policy,
+            ..
```
Geal: if there are more fields in the builder, they should be used, not silently ignored

src/consumer.rs
```diff
@@ -372,17 +411,18 @@ enum InnerConsumer<T: DeserializeMessage, Exe: Executor> {
...
 type MessageIdDataReceiver = mpsc::Receiver<Result<(proto::MessageIdData, Payload), Error>>;
 
+// this is entirely public for use in reader.rs
 pub(crate) struct TopicConsumer<T: DeserializeMessage, Exe: Executor> {
-    consumer_id: u64,
-    config: ConsumerConfig,
-    topic: String,
-    messages: Pin<Box<MessageIdDataReceiver>>,
-    engine_tx: mpsc::UnboundedSender<EngineMessage<Exe>>,
+    pub(crate) consumer_id: u64,
```
Geal, line 416: do we need all of those to be pub(crate)?

Geal (changes requested):

.gitignore
Keksoj: done

README.md
Keksoj: done

examples/reader.rs
Keksoj: sure, you're right. it's done

src/client.rs
Keksoj: done.

src/client.rs
Keksoj: done!

src/consumer.rs
Keksoj: That may be true, but they are usefull in the reader examples, while remaining optional and leaving backward compatibility. It would be as practical to just deliver them in this PR.

Geal: ok

src/consumer.rs
Keksoj: as a matter of fact, no, we don't. I ran `cargo check`ed all of them, only `consumer_id`, `config`, and `dead_letter_policy` are needed elsewhere.

src/consumer.rs
Keksoj: done with `unused_variables` macro

Geal:
thanks!

PR #165, opened 2021-06-24
Support insecure TLS connections

fantapsody:
Add an option to allow insecure TLS connections which is useful in test environments.

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -75,6 +75,9 @@ impl std::default::Default for OperationRetryOptions {
 pub struct TlsOptions {
     /// contains a list of PEM encoded certificates
     pub certificate_chain: Option<Vec<u8>>,
+
+    /// Allow insecure TLS connection if set to true
+    pub allow_insecure_connection: bool
 }
 
 enum ConnectionStatus<Exe: Executor> {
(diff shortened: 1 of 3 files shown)


src/connection_manager.rs
```diff
@@ -75,6 +75,9 @@ impl std::default::Default for OperationRetryOptions {
 pub struct TlsOptions {
     /// contains a list of PEM encoded certificates
     pub certificate_chain: Option<Vec<u8>>,
+
+    /// Allow insecure TLS connection if set to true
+    pub allow_insecure_connection: bool
```
freeznet: can we have one more flag like `validate_hostname`?

fantapsody: Could you explain a bit more?

freeznet: @fantapsody i noticed that you set both `danger_accept_invalid_hostnames` and `danger_accept_invalid_certs` according to `allow_insecure_connection`. My suggestion is to use 2 options, like `validate_hostname` and `allow_insecure_connection`, so that it will act like other pulsar client.

fantapsody: Thanks, I see, there is a configuration `isTlsHostnameVerificationEnable` in the Java pulsar cluster.

freeznet (approved):

PR #168, opened 2021-08-12
fix: Lock nom version to 7.0.0-alpha1

omegaphoenix:
7.0.0-alpha2 removes the `named!` macro.

Addresses https://github.com/wyyerd/pulsar-rs/issues/167

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ travis-ci = {repository = "wyyerd/pulsar-rs"}
 bytes = "1.0.0"
 crc = "2.0.0"
 futures = "0.3"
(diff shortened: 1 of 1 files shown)


omegaphoenix:
@Geal Would you have time to approve this MR so that projects adding pulsar-rs or upgrading can compile without locking the nom version please?  (Previously, since nom wasn't pinned, the breaking patch update to 7.0.0-alpha2 was picked up which requires other projects which use pulsar-rs to pin the nom version.)

Alternatively, I would be happy to help upgrade to the latest version of nom (or downgrade to the last stable version).

juanviamonte:
Thanks for this PR @omegaphoenix

Geal:
Pulsar 4.1 is now released with nom 7.0: fb04666

PR #172, opened 2021-09-16
Add topic to consumer close error logs

sammoh94:
When inspecting broker logs, This will make it easier to debug which topics are seeing this error


Geal:
Thanks

Issue #148, opened 2021-03-09
Documentation


Geal:
we are still missing a lot of documentation, on method, fields, etc, and code examples along with them.

add
```
#![warn(missing_docs)] 
#![warn(missing_doc_code_examples)]
```

to src/lib.rs to see how much

Keksoj:
At the very least we need to update the code example with `Cargo.toml` imports, even if commented out, for the average user to know how to include `serde` with `derive` macros enabled, and the like.

Also, there is some `extern crate` syntax in the docs that need to get out.

Keksoj:
@Geal would you please add me as a contributor to the project? So I can push a branch directly instead of forking? Is the `master` branch protected? This would be a must-have.

Geal:
I don't have the rights for that. @stearnsc could you give access to @Keksoj ?
He's from Clever Cloud too, and since I left the company, he's taking over my work on Pulsar, since I have less time to work on it at my new job (maybe after a while I'll convince them to use pulsar :laughing: )

Geal:
@Keksoj usually the policy was to fork under @clevercloud for repositories we do not have rights to

Geal:
@Keksoj in the meantime, if you need something to be merged, ping me :)

PR #179, opened 2021-11-08
Clearer examples in the doc

Keksoj:
As pointed out in 

- #148 

The doc is lacking.

This PR is here to at least fix the doc's wacky examples. Copy-pasting the examples doesn't work today, because of implicit and unspecified dependencies.


Keksoj:
Ping @Geal ;-)

Geal:
Great!

PR #181, opened 2021-11-19
Update connection.rs

18601673727:
https://github.com/wyyerd/pulsar-rs/issues/163


fantapsody (changes requested):
Simply accepting the `http` scheme doesn't make the connection support `HTTP` protocol automatically, it should still use the pulsar binary protocol. I think we need to implement a separate procedure for HTTP discovery.

tisonkun:
Closed as invalid. Even if we want to support talk to the HTTP endpoint of a Pulsar server, we need a design and implementation quite different from this patch.

PR #182, opened 2021-11-29
Support oauth2 authentication

fantapsody:
This PR adds support of authentication methods that generate authentication data dynamically.

- Add a trait `authentication::Authentication` to abstract authentication providers
- Implement `authentication::Authentication` for `connection::Authentication` to make it as a special `authentication::Authentication` implementation.
- Refactor code to use `authentication::Authentication`
  - Initialize the authentication provider at the time of creating the `ConnectionManager`
  - Get the authentication data asynchronously after connected to the remote server
- Implement a token provider
- Implement an (optional) oauth2 provider
  - Basically follows the example of the oauth2 implementation in the pulsar client: https://github.com/apache/pulsar/tree/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/auth/oauth2
  - Cache the access token in memory and request new tokens automatically.
- Add oauth2 support in examples of producer and consumer.

### DIFF
--- a/src/authentication.rs
+++ b/src/authentication.rs
@@ -80,13 +80,13 @@ pub mod oauth2 {
     pub struct OAuth2Params {
         pub issuer_url: String,
         pub credentials_url: String,
-        pub audience: String,
+        pub audience: Option<String>,
         pub scope: Option<String>,
     }
 
     impl Display for OAuth2Params {
         fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-            write!(f, "OAuth2Params({}, {}, {}, {:?})", self.issuer_url, self.credentials_url, self.audience, self.scope)
+            write!(f, "OAuth2Params({}, {}, {:?}, {:?})", self.issuer_url, self.credentials_url, self.audience, self.scope)
         }
     }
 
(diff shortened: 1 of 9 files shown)


fantapsody:
Hi @Geal @stearnsc , could you take a look at the PR?

freeznet (approved):
LGTM

src/authentication.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    use crate::authentication::Authentication;
+    use crate::error::AuthenticationError;
+
+    #[derive(Deserialize, Debug)]
+    struct OAuth2PrivateParams {
+        client_id: String,
+        client_secret: String,
+        client_email: Option<String>,
+        issuer_url: Option<String>,
```
EronWright, line 76: As a stretch goal, please prefer the `issuer_url` within the json file over the configured `issuer_url`.  This makes it possible to support the use of numerous issuers (e.g. for migration purposes).

src/authentication.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    #[derive(Deserialize, Debug)]
+    pub struct OAuth2Params {
+        pub issuer_url: String,
+        pub credentials_url: String,
+        pub audience: String,
```
EronWright: Please treat `audience` as optional, since it is an Auth0-specific extension.   Azure, in contrast, uses a scope for the same purpose.

Issue #184, opened 2021-12-13
Can `Reader` implement `Send`?


shanicky:
Hi, wyyerd!

We are recently using this amazing client

We want to use `tokio` in combination with `Reader` for consumption, we expect to use `tokio::spawn` to spawn tokio threads for low level parallel consumption, but this requires `Reader` to implement `Send` trait.

We are curious if it is possible to add `Send` support to `Reader`, thanks!

PR #185, opened 2021-12-13
make Reader Sync + Send

shanicky:
Hi

We are using `Reader` interface recently and we want to generate tokio thread for consumption via `tokio::spawn`, but this requires Reader to be `Sync + Send`, so this is a patch for issue #184


stearnsc (approved):

Keksoj:
well if it works on your end, this looks OK to me. What do you think @Geal  ?

PR #186, opened 2021-12-17
Add get_consumer_stats function

dylankyc:
Add get consumer stats functionality to check if any consumers go wrong, we can do something like reconnecting brokers, when wired things happen in the pulsar server side.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -341,6 +341,18 @@ impl<Exe: Executor> ConnectionSender<Exe> {
         .await
     }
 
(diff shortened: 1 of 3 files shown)


dylankyc:
Ping @Geal ;-)

stearnsc (approved):
LGTM

DonghunLouisLee (approved):

DonghunLouisLee:
@lichuan6  Hi lichuan! would it be possible to fetch from upstream and resolve conflicts? 

Thanks

PR #191, opened 2022-02-27
Support LookupResponse without broker_url

fantapsody:
Currently, the client returns `ServiceDiscovery(NotFound)` error when there is no `broker_url` in the lookup response. However, there're cases that only `broker_tls_url` is available, the client should handle this case properly.


EronWright (approved):
Seems fine but wasn't totally clear to me whether the scheme of the service url is what's used for the broker url.  As the other language bindings do.

PR #216, opened 2022-06-17
Fix thread leak

### LINKED ISSUE #212: Possible connection executor tasks overflow
The client's sender seems to create a task for each received proto message and never checks if the future finishes.

https://github.com/streamnative/pulsar-rs/blob/c31fc8335e4e321c85e0024aa4189c318a581082/src/client.rs#L560

If a reconnection loop is ongoing then all the messages are stacked in tasks without any timeout or cancellation which leads to task overflow.

KannarFr:
Closes https://github.com/streamnative/pulsar-rs/issues/212

I'm not a rust developer, let's consider this PR as try to fix. Feedbacks appreciated.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -8,14 +8,18 @@ use crate::connection_manager::{
     BrokerAddress, ConnectionManager, ConnectionRetryOptions, OperationRetryOptions, TlsOptions,
 };
 use crate::consumer::{ConsumerBuilder, ConsumerOptions, InitialPosition};
+use crate::error::ConnectionError;
 use crate::error::Error;
 use crate::executor::Executor;
 use crate::message::proto::{self, CommandSendReceipt};
 use crate::message::Payload;
 use crate::producer::{self, ProducerBuilder, SendFuture};
 use crate::service_discovery::ServiceDiscovery;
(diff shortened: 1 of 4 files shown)


src/client.rs
```diff
@@ -555,9 +562,27 @@ async fn run_producer<Exe: Executor>(
...
-            Ok(future) => {
+            Ok(send_f) => {
+                let delay_f = client
+                    .executor
+                    .delay(client.operation_retry_options.operation_timeout);
+
                 let _ = client.executor.spawn(Box::pin(async move {
-                    let _ = resolver.send(future.await);
+                    pin_mut!(delay_f);
+                    pin_mut!(send_f);
```
DonghunLouisLee, line 572: Is there a reason why send_f should also be pinned?

DonghunLouisLee:
Since this #217 includes this PR, I'll close this issue and review #217 instead. But this is a very good point though!

src/client.rs
DonghunLouisLee: My understanding is that since send_f: SendFuture has already been moved in line 570, when select returns with Either::Right(), the send_f will be dropped, thus pin_mut is unnecessary.

PR #218, opened 2022-07-13
Update zstd version to support Solana crates

n00m4d:
## What
Update zstd version 0.10 -> 0.11
## Why
To support Solana crates


sheldon-metaplex:
+1 this is something we need. Not sure if it has other impacts on the code base, though.

omegaphoenix:
[actix-web 4.1.0](https://github.com/actix/actix-web/blob/4bbe60b609ef1d66475b26fab9476a2fbcf6fc45/actix-web/Cargo.toml#L122) also requires zstd 0.11

PR #222, opened 2022-08-19
bump prost to 0.11.0

skyzh:
As title. This is a breaking change and also needs to bump and release new version of pulsar-rs. Not sure whether this PR is the best way to bump dependencies. Thanks for reviewing!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,6 +18,11 @@ jobs:
       - name: Start Pulsar Standalone Container
         run: docker run --name pulsar -p 6650:6650 -p 8080:8080 -d -e GITHUB_ACTIONS=true -e CI=true streamnative/pulsar:${{ matrix.pulsar-version }} /pulsar/bin/pulsar standalone
 
+      - uses: arduino/setup-protoc@v1
+        name: Install protoc
+        with:
+          version: "3.x"
+
       - uses: actions/checkout@v2
       - name: Build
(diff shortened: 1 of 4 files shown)


fantapsody:
Could you explain the motivation of the PR? I checked the release note of https://github.com/tokio-rs/prost/releases/tag/v0.11.0 and it didn't mention the version has a breaking change. @skyzh

skyzh:
Generally bumping prost would need to bump major semantic version of projects depending on it to keep APIs consistent and semantic version convention. It would also be okay to only bump a minor version for the pulsar package if we don’t need to follow semantic version convention. This practice is followed by a lot of projects like tonic.

skyzh:
… and Prost’s release note mentions that “Relase 0.11 brings a few breaking changes and changes to how prost sources protoc”, though the change looks minor.

skyzh:
Need to install protoc in CI manually. Will fix CI scripts.

skyzh:
Need approval for workflow. I believe it should work for now.

skyzh:
ready for review

tisonkun:
@skyzh since it requires a major version bump, I'm curious about the motivation of bumping prost version itself.

If we don't see a requirement yet, we can postpone this version bump.

skyzh:
That would be fine, we can do this later.

FlorentinDUBOIS:
Hi, this bump has been done in the pr #233. Thank you for your contribution :)

PR #227, opened 2022-08-29
fix: Using a weak reference in producers closing callback

CleverAkanoa:
**Description**

We're observing some connection that never drop even when the connection is down.

Our assumption is that the Arc::clone keep the reference alive inside the callback.


KannarFr:
@CleverAkanoa can you remove the draft tag?

FlorentinDUBOIS (approved):

freeznet (approved):

tisonkun (approved):
Thanks for your contribution. LGTM.

Merging...

PR #233, opened 2022-09-28
Implements tracing capabilities through telemetry compiler feature flag

FlorentinDUBOIS:
Hello :wave:, 

To find an issue on production, I needed more observability of the software that use this crate, to get more insights, I put tracing annotation on functions behind a compiler feature flag to not add change to the current behavior of the driver.

Besides, I saw that the driver needed some love. So, I have updated all dependencies using the same way to write the semantic version using `^x.x.x` which allow to lock easily a version, if something does not work as expected in future version of a crate through this syntax `=x.x.x`. I also have updated the Rust edition which is now the latest one `2021`, which asked at least Rust `1.56.0`, if my memory is good.

Last thing, I have set the right version number of the crate on the `Cargo.toml` according to the latest git tag.

Have a wonderful day!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ openidconnect = { version = "^2.3.2", optional = true }
 oauth2 = { version = "^4.2.3", optional = true }
 serde = { version = "^1.0.145", features = ["derive"], optional = true }
 serde_json = { version = "^1.0.85", optional = true }
+tracing = { version = "^0.1.36", optional = true }
 async-trait = "^0.1.57"
 data-url = { version = "^0.2.0", optional = true }
 uuid = {version = "^1.1.2", features = ["v4", "fast-rng"] }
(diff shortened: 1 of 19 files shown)


FlorentinDUBOIS:
Besides, if this pull reques is merged, it closes #222

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "pulsar"
-version = "4.1.1"
-edition = "2018"
+version = "4.1.3"
+edition = "2021"
```
KannarFr, line 4: these should be done in the release commit after merging.

KannarFr (changes requested):
Just a strange version change in a PR :p, otherwise LGTM.

Cargo.toml
CleverAkanoa: I agree

FlorentinDUBOIS: I agree, but the version 4.1.3 is already published according to git tag. I will create another commit to do this change.

FlorentinDUBOIS: I have remove the commit that do this change

FlorentinDUBOIS:
@KannarFr @CleverAkanoa could you approve if it is ok for you?
@utahkay @vmalloc do you agree with those changes?

vmalloc:
Not sure I'm the right person to ask...

utahkay:
I'm not sure I'm the right person to ask, either...

CleverAkanoa (approved):
Seems good for me.

Maybe we can document this telemetry feature in the Readme file.

FlorentinDUBOIS:
You are right, I will do that 😇

Issue #189, opened 2022-02-14
Consumer gets stuck indefinitely in some cases where Pulsar/ZK restarts


vmalloc:
Somewhat related to #164, I'm experiencing a similar hang but on the consumer side. Using the pulsar consumer as a `Stream`, it seems like after certain occasions in which ZK nodes fail (in my case after experiencing errors), the pulsar cluster recovers just fine, but the client becomes stuck, hanging the entire task that called `next()`. This is especially severe because running `try_next` in a `select!` loop like I do actually calls the entire select branch to hang, meaning the reactive loop stops responding to waking futures.

The logs don't say much except for these lines that get repeated several times around the time of the hang:
```
ERROR pulsar::consumer] Consumer: messages::next: returning Disconnected
INFO  pulsar::connection_manager] Connected n°5991518907389693436 to pulsar-broker-0.pulsar-broker.pulsar.svc.cluster.local:6650 via proxy pulsar://pulsar-proxy.pulsar.svc in 78ms
ERROR pulsar::connection_manager] could not ping connection -7903802395670970975 to the server at pulsar://pulsar-proxy.pulsar.svc: Disconnected
```

Restarting the service altogether (causing a complete reconnect) seems to have fixed it completely, so there must be a bug in the reconnection process here I guess....

EDIT: I'm using the latest version of the client library (4.1.1)

vmalloc:
FYI @Geal , I managed to reproduce this very easily. Having a single producer and a single consumer against a locally running pulsar (I used https://github.com/vmalloc/pulsar-cli to do that):

1. Start both consumer and producer
2. Kill the pulsar instance with SIGKILL (may need to try a couple of times before it reproduces
3. After a very small number of attempts, the consumer emits the following logs:
```
[2022-02-15T10:28:06Z DEBUG pulsar::connection] Connecting to pulsar://127.0.0.1: 127.0.0.1:6650
[2022-02-15T10:28:06Z DEBUG pulsar::consumer] consumer engine stopped: Err(ServiceDiscovery(Connection(Io(Os { code: 61, kind: ConnectionRefused, message: "Connection refused" }))))
[2022-02-15T10:28:06Z ERROR pulsar::consumer] could not close consumer Some("pulsar-cli")(14550240535650196156) for topic test2: Disconnected
```

After that, the consumer stream continuously yields `None` as the message, even though the server recovers pretty quickly.

vmalloc:
[...]

PR #234, opened 2022-09-30
fix: improve refresh mechanism in `MultiTopicConsumer::poll_next`

kagetsuki1997:
Hi, the problem I met was `MultiTopicConsumer::next` will pending forever if pulsar was disconnected. I think it is similar to #189

I found out that `MultiTopicConsumer::poll_next` will remove `topic` and `consumer` if it received None or Err from consumer. This mechanism will cause `MultiTopicConsumer::poll_next` always pending.

There is a refresh mechanism in `MultiTopicConsumer::poll_next` which will call `MultiTopicConsumer::update_topics` every 30secs in default. But in `MultiTopicConsumer::update_topics` , it only try to reconnet the topics which match `MultiTopicConsumer::topic_regex`.

So I think if we recreate the consumers of `MultiTopicConsumer::topics` in `MultiTopicConsumer::update_topics`. It will solve the problem.

Thanks.

By the way, it is the first PR I create to an open source project :D

### DIFF
--- a/src/consumer.rs
+++ b/src/consumer.rs
@@ -268,11 +268,10 @@ impl<T: DeserializeMessage, Exe: Executor> Consumer<T, Exe> {
                 }))
                 .await?;
 
-                let consumers: BTreeMap<_, _> = consumers
-                    .into_iter()
-                    .map(|c| (c.topic(), Box::pin(c)))
-                    .collect();
-                let topics = consumers.keys().cloned().collect();
+                let consumers: BTreeMap<_, _> =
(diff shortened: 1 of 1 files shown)


FlorentinDUBOIS:
Hey, Thanks you ! Glad to meet you there for your first contribution ! I will test your pull request on Clever-Cloud staging, I will keep you updated of behaviors of your pull request, we have a reproducible environment for the issue #189. So, I could be able to test your case. I hope to merge it, to be able to embed it into the next release.

Issue #224, opened 2022-08-20
Proper Version Control


DonghunLouisLee:
Since pulsar-rs currently does not have proper version control policies, I suggest we move forward with following policies as below

1. Major Update(1.x.x -> 2.x.x) will follow pulsar major version update(2.x.x -> 3.x.x) 
2. Minor Update(2.1.x -> 2.2.x) will follow pulsar minor version update(2.1.x -> 2.2.x) 
3. Patch Update(2.1.1 -> 2.1.2) will be a collection of necessary updates

For now, since pulsar-rs is INCOMPLETE as it does not support all the features, it would be a good idea to set a target pulsar version that this client library aims to match.

I propose 2.10.x as the target apache version and due date as 2022 Decemeber. Until then, all new features beyond 2.10.x should not be prioritized and PRs should be focused on solving existing bugs and features within 2.10.x. 

Once our goal is satisfied, It would be a good idea to follow version control as mentined above though we would have to start from 4.10.x which would be compatible with apache pulsar 2.10.x since this crate has already reached version 4.1.2 as of 2022/08 

Following is list of open issues that should be tackled for version fix(list will be updated continuously). Contributors for this crate are strongly encouraged to handle below issues. 

1. Authentication 
    * [x] #221 
    * [ ] #197
2. Producer 
    * [ ] #104
    * [x] #213
3. Consumer
    * [ ] #194
    * [ ] #189
4. Misc
    * [ ] #171

DonghunLouisLee:
I'll first handle existing PRs then write down action items(features that are missing)

tisonkun:
Thanks for starting this discussion @DonghunLouisLee! 

Generally, client libraries should be backward compatible during a major version. And I can see that the `pulsar` crate already takes versions from `0.x` to `4.x`, including `1.x`, `2.x`, and `3.x`.

I'd prefer to continue the version strategy that releasing the `pulsar` crate standalone while holds a compatibility matrix as other client libraries do:

* [pulsar-client-node](https://github.com/apache/pulsar-client-node#compatibility)
* [ruby-kafka](https://github.com/zendesk/ruby-kafka#compatibility)

Except Pulsar bumps its major version, its protocol should keep backward compatible. Thus, we don't have to force the rust client to do simultaneous releases with the main repo.

cc @codelipenghui

tisonkun:
[...]

PR #236, opened 2022-10-11
connection: try all addresses on connecting

tisonkun:
This closes #134.

Signed-off-by: tison <wander4096@gmail.com>
Co-authored-by: Marc-Antoine Perennou <Marc-Antoine@Perennou.com>

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-09-08"
+channel = "stable"
 components = ["rustfmt", "clippy"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "nightly-2022-09-08"
+components = ["rustfmt", "clippy"]
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -788,12 +788,18 @@ impl<Exe: Executor> Connection<Exe> {
             return Ok(Connection { id, url, sender })
         }
 
-        let retryable_errors = errors
-            .drain_filter(|e| e.establish_retryable())
-            .collect::<Vec<ConnectionError>>();
+        let mut fatal_errors = vec![];
+        let mut retryable_errors = vec![];
+        for e in errors.into_iter() {
+            if e.establish_retryable() {
+                retryable_errors.push(e);
+            } else {
+                fatal_errors.push(e);
+            }
+        }
 
         return if retryable_errors.is_empty() {
-            error!("connection error, not retryable: {:?}", errors);
+            error!("connection error, not retryable: {:?}", fatal_errors);
             Err(ConnectionError::Io(std::io::Error::new(
                 std::io::ErrorKind::Other,
                 "fatal error when connecting to the Pulsar server",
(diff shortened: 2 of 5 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly-2022-09-08"
```
tisonkun: I don't find documents about MSRV so just assume that we use the nightly version as usual.

Add this file as it's required to enable feature gates (`drain_filter` in this patch). Although it's possible to write a utility function, if we can just use stdlib, I'd prefer to use it.

tisonkun:
ping @FlorentinDUBOIS

rust-toolchain.toml
FlorentinDUBOIS: To me, the toolchain have to be stable, if it is not set. I am strongly discourage nightly requirement on production driver.

src/connection.rs
```diff
@@ -733,56 +735,73 @@ impl<Exe: Executor> Connection<Exe> {
...
+                    continue;
+                }
+            };
 
-        Ok(Connection { id, url, sender })
+            return Ok(Connection { id, url, sender })
+        }
+
+        let retryable_errors = errors
+            .drain_filter(|e| e.establish_retryable())
```
FlorentinDUBOIS: To me, this line should be written like below to remove the nightly requirement:
```suggestion
            .drain()
            .filter(|e| e.establish_retryable())
```

FlorentinDUBOIS (changes requested):
Thank for the pull request, but I cannot merge since it requires nightly rust build.

tisonkun:
@FlorentinDUBOIS OK. I will point the toolchain to stable and find an alternative to the related code.

FlorentinDUBOIS:
Thank you! I provided you a way to do it, I hope it will help you :)

FlorentinDUBOIS:
Once, your modifications will be merge, I am quite motivated to release a 4.2.0

src/connection.rs
tisonkun: Not exactly. `drain` take a range instead of a condition. And if we move the original errors, we cannot use it in the line:

```rust
error!("connection error, not retryable: {:?}", errors);
```

But perhaps we don't have to do error handling so fine-grained.

tisonkun: But we can simply write some imperative code here. I'm preparing a commit on this.

tisonkun:
@FlorentinDUBOIS fixup pushed and tested locally. Let's wait for the CI verification ;)

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-09-08"
+channel = "stable"
```
tisonkun, line 2: @FlorentinDUBOIS Now it means the latest stable version. I think it's OK for now and we may choose an fixed MSRV when #224 is finished and we think we've completed the major feature set.

FlorentinDUBOIS (approved):

rust-toolchain.toml
FlorentinDUBOIS: I am ok with that, but nightly is really tricky :sweat_smile:

PR #237, opened 2022-10-12
ci: remove redundant steps

tisonkun:
IIRC the default runner bundles protoc executable.


tisonkun:
Well. I think it's previous version of `tonic-build` provides the bundled `protoc`.

PR #240, opened 2022-11-15
feat: support configure compression level

### LINKED ISSUE #239: Add control over compression level
Hi!

Now the library allows enabling compression before sending to Pulsar but does not allow to choose compression level - the defaults are used. Could you please add the possibility to control compression level for the compression algorithms? It will allow to configure more precisely tradeoff between CPU and size of the messages.

Thanks in advance!

tisonkun:
This closes https://github.com/streamnative/pulsar-rs/issues/239.

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -16,9 +16,9 @@ pub struct CompressionLz4 {
     pub mode: CompressionMode,
 }
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Default, Clone, Copy, Debug)]
 pub struct CompressionZlib {
-    pub level: u32,
+    pub level: flate2::Compression,
 }
 
 #[derive(Clone, Copy, Debug)]
@@ -36,14 +36,6 @@ impl Default for CompressionLz4 {
     fn default() -> Self {
         CompressionLz4 {
             mode: CompressionMode::DEFAULT,
-        }
-    }
-}
-
-impl Default for CompressionZlib {
-    fn default() -> Self {
-        CompressionZlib {
-            level: flate2::Compression::default().level(),
         }
     }
 }
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -1,0 +1,56 @@
+//! Compression strategy configs
+
+#[derive(Clone, Debug)]
+pub enum Compression {
+    None,
+    Lz4(CompressionLz4),
+    Zlib(CompressionZlib),
+    Zstd(CompressionZstd),
+    Snappy,
+}
+
+#[derive(Clone, Debug)]
+pub struct CompressionLz4 {
+    pub mode: lz4::block::CompressionMode,
+}
+
+#[derive(Clone, Debug)]
+pub struct CompressionZlib {
+    pub level: u32,
+}
+
+#[derive(Clone, Debug)]
+pub struct CompressionZstd {
+    pub level: i32,
+}
+
+impl Default for Compression {
+    fn default() -> Self {
+        Compression::None
+    }
+}
+
+impl Default for CompressionLz4 {
+    fn default() -> Self {
+        CompressionLz4 {
+            mode: lz4::block::CompressionMode::DEFAULT,
+        }
+    }
+}
+
+impl Default for CompressionZlib {
+    fn default() -> Self {
+        CompressionZlib {
+            level: flate2::Compression::default().level(),
+        }
+    }
+}
+
+impl Default for CompressionZstd {
+    fn default() -> Self {
+        CompressionZstd {
+            level: zstd::DEFAULT_COMPRESSION_LEVEL,
+        }
+    }
+}
+
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -3,66 +3,44 @@
 use lz4::block::CompressionMode;
 
 /// Wrapper of supported compression algorithms
-#[derive(Clone, Debug)]
+#[derive(Default, Debug)]
 pub enum Compression {
+    #[default]
     None,
-    Lz4(CompressionLz4),
-    Zlib(CompressionZlib),
-    Zstd(CompressionZstd),
+    /// Options of the [lz4](https://lz4.github.io/lz4/) algorithm
+    Lz4 {
+        /// compression mode of lz4 to be used
+        mode: CompressionMode
+    },
+    /// Options of the [zlib](https://www.zlib.net/) algorithm
+    Zlib {
+        /// compression level of zlib to be used (0-9)
+        level: flate2::Compression,
+    },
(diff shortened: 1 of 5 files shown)


tisonkun:
According to https://github.com/streamnative/pulsar-rs/issues/239#issuecomment-1315651065, pending to add a compression abstraction to hold these configs.

tisonkun:
Updated.

cc @zamazan4ik @FlorentinDUBOIS @DonghunLouisLee 

Since I change the type of `compression` config key, it's possible that we need a major version bump.

src/producer.rs
```diff
@@ -422,26 +423,26 @@ impl<Exe: Executor> TopicProducer<Exe> {
 
         let topic = topic.clone();
         let batch_size = options.batch_size;
-        let compression = options.compression;
+        let compression = options.compression.clone();
```
tisonkun, line 426: This should be `Copy`. Blocked by upstream https://github.com/10XGenomics/lz4-rs/pull/30.

zamazan4ik:
FMPOV looks fine. Thank you for the PR!

tisonkun:
@zamazan4ik do you have a suggestion on which version to release including this patch? 4.3.0 or 5.0.0.

Semantically we should release 5.0.0 but actually this crate is not stable yet I would say (refer #224).

zamazan4ik:
@zamazan4ik No, I have no. For me personally, there is no considerable difference :) So I suggest just bumping the version to 4.3.0 .

src/compression.rs
```diff
@@ -0,0 +1,68 @@
+//! Compression strategy configs
+
+use lz4::block::CompressionMode;
+
+/// Wrapper of supported compression algorithms
+#[derive(Clone, Debug)]
+pub enum Compression {
```
FlorentinDUBOIS, line 5: I am not sure if subtypes are allowing this, but to me the enum should be written like this:

```rust
# [derive(Clone, Debug, Default)]
pub enum Compression {
  # [default]
  None,
  Lz4 { mode: CompressionMode },
  Zlib { level: flate2::Compression },
  Zstd { level: i32 },
  Snappy,
}
```

Writing the enum like this will reduce a lot the code source as we do not need to create intermediate structures.

FlorentinDUBOIS (changes requested):

src/compression.rs
tisonkun: Updated.

src/compression.rs
```diff
@@ -0,0 +1,46 @@
+//! Compression strategy configs
+
+use lz4::block::CompressionMode;
+
+/// Wrapper of supported compression algorithms
+#[derive(Default, Debug)]
+pub enum Compression {
```
fantapsody, line 5: Should the availability of compression features be checked here?

src/producer.rs
```diff
@@ -127,7 +128,7 @@ pub struct ProducerOptions {
     /// batch message size
     pub batch_size: Option<u32>,
     /// algorithm used to compress the messages
-    pub compression: Option<proto::CompressionType>,
+    pub compression: Option<Compression>,
```
fantapsody, line 131: This looks like a breaking change, is it possible to add a new field and deprecate this one instead of changing its type directly?

tisonkun: With this comment, I tend to revert 34bf64ae7edb318b9651d89698b9587ffab2f7bb which applies @FlorentinDUBOIS's comment.

For breaking change, I make it with these trade-off considerations:

1. `compression` is the best name for this config, if we keep b/w compatibility, we need to choose a new name like `compression_mode` or `compression_strategy`.
2. This library - although the latest release is 4.2.0 - is still incomplete for production and I think exposing `proto::*` is a temporary solution.
3. If - with 34bf64ae7edb318b9651d89698b9587ffab2f7bb reverted - we use a wrapper struct to this field, and as long as we implement `Default` for it, it can be b/w compatible from now on.
4. Deprecating and fallback logics are also overhead/debt.

src/compression.rs
tisonkun: OK. I get your point to make something like conditional compilation. Let me try it out.

tisonkun:
Updated. cc @fantapsody @FlorentinDUBOIS

src/producer.rs
tisonkun: Fixed in https://github.com/streamnative/pulsar-rs/pull/240/commits/c889a08fd9aa34cf6aab74c3e1d09c67bca77b1e

src/compression.rs
tisonkun: Fixed in https://github.com/streamnative/pulsar-rs/pull/240/commits/c889a08fd9aa34cf6aab74c3e1d09c67bca77b1e

src/producer.rs
```diff
@@ -422,27 +423,7 @@ impl<Exe: Executor> TopicProducer<Exe> {
 
         let topic = topic.clone();
         let batch_size = options.batch_size;
-        let compression = options.compression;
-
-        match compression {
```
tisonkun, line 427: Now the compression modes are structs under control. We exclude the corresponding enum members if features are unset. Thus, we're sure that there's no config-implementation mismatch.

fantapsody (approved):

src/producer.rs
fantapsody: I see. Maybe we can implement the `From<proto::CompressionType >` trait for `Compression` to make the code migration easier.

tisonkun: Good point. Let me try it out.

tisonkun: @fantapsody how can you imagine it eases the case?

Perhaps `compression: Some(proto::CompressionType::Snappy.into())`? Since it also touches the code anyway, I think it's not quite better than `compression: Some(Compression::Snappy(CompressionSnappy::default()))` with `use compression::...`.

The problem here is, `proto::CompressionType` is available always, and if we support such conversion, we need to play some runtime check which we removed at https://github.com/streamnative/pulsar-rs/pull/240#discussion_r1027263857 again. I don't think it's a valuable trade-off.

tisonkun:
Pending to merge by the end of this week.

I'm going to release a 5.0.0 for breaking changes. @fantapsody @FlorentinDUBOIS @DonghunLouisLee what do you think?

For the increasing version number, the only risk I can see is that they're all beyond 1.0.0 but as stated in https://github.com/streamnative/pulsar-rs/issues/224, they're actually untested for compatibility with a specific official Pulsar version.

FlorentinDUBOIS (approved):

tisonkun:
Pending to merge tomorrow and stage as 5.0.0.

Perhaps I'll add rustfmt/clippy tools to reformat code stably before calling the release.

PR #243, opened 2022-12-01
refactor: Split consumer.rs file

CleverAkanoa:
This PR aims to ease the code navigability of the consumer part.

consumer.rs involves more than 2000 lines of codes, it's quite hard to know where you are.

Even all structure are related to consumer, they can be split in more than one files.

No code has been changed, only moved and visibility opened when it was mandatory.


FlorentinDUBOIS (approved):

KannarFr:
So much easier to read! <3.

tisonkun:
@CleverAkanoa is this PR ready for review now?

CleverAkanoa:
I think so :)

src/consumer/message.rs
```diff
@@ -0,0 +1,48 @@
...
+///
+/// it is generic over the type it can be deserialized to
+pub struct Message<T> {
+    /// origin topic of the message
+    pub topic: String,
+    /// contains the message's data and other metadata
+    pub payload: Payload,
+    /// contains the message's id and batch size data
+    pub message_id: MessageData,
+    pub(crate) _phantom: PhantomData<T>,
```
tisonkun:
```suggestion
    pub(super) _phantom: PhantomData<T>,
```

Narrow the scope as possible. Previously it was private.

src/consumer/topic.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub(crate) consumer_id: u64,
+    pub(crate) config: ConsumerConfig,
+    topic: String,
+    messages: Pin<Box<MessageIdDataReceiver>>,
+    engine_tx: mpsc::UnboundedSender<EngineMessage<Exe>>,
+    #[allow(unused)]
+    data_type: PhantomData<fn(Payload) -> T::Output>,
+    pub(crate) dead_letter_policy: Option<DeadLetterPolicy>,
+    pub(crate) last_message_received: Option<DateTime<Utc>>,
+    pub(crate) messages_received: u64,
```
tisonkun:
```suggestion
    pub(super) last_message_received: Option<DateTime<Utc>>,
    pub(super) messages_received: u64,
```

ditto

src/consumer/topic.rs
```diff
@@ -0,0 +1,374 @@
...
+}
+
+impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
+    #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    pub async fn new(
```
tisonkun:
```suggestion
    pub(super) async fn new(
```

tisonkun (commented):
Thanks for your effort! Minor comments inline.

tisonkun (approved):
Thanks for your contribution @CleverAkanoa!

Merging...

CleverAkanoa:
Thanks @tisonkun <3

src/consumer/multi.rs
```diff
@@ -0,0 +1,338 @@
+use std::{
+    collections::{BTreeMap, VecDeque},
+    fmt::Debug,
+    future::Future,
+    iter,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use async_std::prelude::Stream;
```
lukesteensen, line 10: This causes a compile error when the `async_std` feature is not active.

src/consumer/topic.rs
```diff
@@ -0,0 +1,374 @@
+use std::{
+    marker::PhantomData,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+    time::{Duration, Instant},
+};
+
+use async_std::prelude::Stream;
```
lukesteensen, line 9: Same here.

lukesteensen (commented):
Hello! I just noticed some errors when we bumped from 5.0.0 to 5.0.1 and tracked them back to this PR. Both due to using `async_std` when we don't have that feature active.

src/consumer/multi.rs
tisonkun: We can use `futures::Stream` which is runtime agnostic. I filed a patch: https://github.com/streamnative/pulsar-rs/pull/250.

Issue #242, opened 2022-11-28
Consumer connection is not cleaned well


KannarFr:
When a consumer has an issue, it tries to reconnect, but I experienced consumer's name is already connected error. This error should be handled and ensured to kill the old connection before trying to reconnect. I already implemented it in producer IIRC.

KannarFr:
We should implement https://github.com/streamnative/pulsar-rs/pull/217/files#diff-f2106858cb19f27eaaa29381deed4563b0809c0cb6adc65e6fa4043dee64362fR819 for consumer reconnection.

KannarFr:
@CleverAkanoa can you handle it?

PR #244, opened 2022-12-07
consumer reconnect

CleverAkanoa:
This PR is related to https://github.com/streamnative/pulsar-rs/issues/242

It's an attempt to clean all reference to old connection before trying reconnection.

I'm not sure about the side effect of these modifications, and I was unable to reproduce the behavior in development environment. 

Even by kicking the test container in the middle of running.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -465,7 +465,6 @@ impl<Exe: Executor> ConnectionSender<Exe> {
     }
 
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
-    /// Try to subscribe a consumer to a given topic
     pub async fn subscribe(
         &self,
         resolver: mpsc::UnboundedSender<Message>,
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -465,6 +465,7 @@ impl<Exe: Executor> ConnectionSender<Exe> {
     }
 
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    /// Try to subscribe a consumer to a given topic
     pub async fn subscribe(
         &self,
         resolver: mpsc::UnboundedSender<Message>,
--- a/src/consumer/engine.rs
+++ b/src/consumer/engine.rs
@@ -1,12 +1,13 @@
 use std::{
     collections::{HashMap, HashSet},
     future::Future,
+    io::ErrorKind,
     sync::Arc,
     time::{Duration, Instant},
 };
 
 use futures::{
-    channel::{mpsc, oneshot},
+    channel::{mpsc, mpsc::UnboundedSender, oneshot},
     SinkExt, StreamExt,
 };
 
@@ -24,7 +25,7 @@ use crate::{
     },
     proto,
     proto::{BaseCommand, CommandCloseConsumer, CommandMessage},
-    Error, Executor, Payload, Pulsar,
+    BrokerAddress, Error, Executor, Payload, Pulsar,
 };
 
 pub struct ConsumerEngine<Exe: Executor> {
@@ -39,14 +40,14 @@ pub struct ConsumerEngine<Exe: Executor> {
     messages_rx: Option<mpsc::UnboundedReceiver<RawMessage>>,
     engine_rx: Option<mpsc::UnboundedReceiver<EngineMessage<Exe>>>,
     event_rx: mpsc::UnboundedReceiver<EngineEvent<Exe>>,
-    event_tx: mpsc::UnboundedSender<EngineEvent<Exe>>,
+    event_tx: UnboundedSender<EngineEvent<Exe>>,
     batch_size: u32,
     remaining_messages: u32,
     unacked_message_redelivery_delay: Option<Duration>,
     unacked_messages: HashMap<MessageIdData, Instant>,
     dead_letter_policy: Option<DeadLetterPolicy>,
     options: ConsumerOptions,
-    _drop_signal: oneshot::Sender<()>,
+    drop_signal: oneshot::Sender<()>,
 }
 
 impl<Exe: Executor> ConsumerEngine<Exe> {
@@ -66,7 +67,7 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
         unacked_message_redelivery_delay: Option<Duration>,
         dead_letter_policy: Option<DeadLetterPolicy>,
         options: ConsumerOptions,
-        _drop_signal: oneshot::Sender<()>,
+        drop_signal: oneshot::Sender<()>,
     ) -> ConsumerEngine<Exe> {
         let (event_tx, event_rx) = mpsc::unbounded();
         ConsumerEngine {
@@ -88,7 +89,7 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
             unacked_messages: HashMap::new(),
             dead_letter_policy,
             options,
-            _drop_signal,
+            drop_signal,
         }
     }
 
@@ -389,7 +390,7 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
                 proto::CompressionType::from_i32(compression).ok_or_else(|| {
                     error!("unknown compression type: {}", compression);
                     Error::Consumer(ConsumerError::Io(std::io::Error::new(
-                        std::io::ErrorKind::Other,
+                        ErrorKind::Other,
                         format!("unknown compression type: {}", compression),
                     )))
                 })?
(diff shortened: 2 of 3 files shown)


KannarFr:
Is it still WIP, or ready to be reviewed?

CleverAkanoa:
WIP and ready to review but not by streamnative, I'd like a first pass by us before

src/consumer.rs
```diff
@@ -574,6 +574,32 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
                         .into());
                     }
                 }
+                Err(ConnectionError::Io(e)) => {
+                    if e.kind() != std::io::ErrorKind::TimedOut {
```
FlorentinDUBOIS: To me, you could also retry in the following case as well:

- ConnectionReset
- ConnectionAborted
- NotConnected
- BrokenPipe
- Interrupted
- UnexpectedEof

See [std::io::ErrorKind](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html)

src/consumer.rs
```diff
@@ -574,6 +574,32 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
...
+                        return Err(ConsumerError::Io(e).into());
+                    } else if operation_retry_options.max_retries.is_none()
```
FlorentinDUBOIS: Could be rewrite with a match, that will remove the unwrap

```rust
match <expr> {
Err(ConnectionError::Io(e)) if matches!(e, ErrorKind::TimedOut | ErrorKind::ConnectionReset | ...) => {
  match operation_retry_options.max_retries {
    None | Some(max_retries) if current_retries < max_retries => { .. }
    _ => { .. }
  }
},

```

src/consumer.rs
```diff
@@ -1369,35 +1436,174 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
...
+                    let dur = (Instant::now() - start).as_secs();
+                    log::info!(
+                        "subscribing({}) success after {} retries over {} seconds",
+                        topic,
+                        *current_retries + 1,
+                        dur
+                    );
+                }
+            }
+            Err(ConnectionError::PulsarError(Some(proto::ServerError::ServiceNotReady), text)) => {
```
FlorentinDUBOIS: Same comment as above, you could do much better with match expression

FlorentinDUBOIS (changes requested):

CleverAkanoa:
I've modified all you've requested. @FlorentinDUBOIS 

Please take another look. :)

src/consumer.rs
```diff
@@ -1369,35 +1451,170 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
...
+                            }
+                        }
+                    }
+                    _ => Err(Err(Error::Connection(ConnectionError::PulsarError(
+                        Some(err),
+                        text,
+                    )))),
+                }
+            }
+            Err(ConnectionError::Io(e)) => {
```
FlorentinDUBOIS: Could be write to reduce a level of indentation like this:
```suggestion
            Err(ConnectionError::Io(e)) if matches!(e.kind(),ErrorKind::BrokenPipe | .. | ErrorKind::UnexpectedEof) => ..
```

FlorentinDUBOIS (changes requested):

src/consumer.rs
Akanoa: Sorry but I don't understand

each indentation level has its particular error message.

What do you to reduce ?

CleverAkanoa:
I've rebase on master since the refactoring 

https://github.com/streamnative/pulsar-rs/pull/243

src/connection.rs
```diff
@@ -465,6 +465,7 @@ impl<Exe: Executor> ConnectionSender<Exe> {
     }
 
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    /// Try to subscribe a consumer to a given topic
```
tisonkun:
```suggestion
```

Not quite valuable.

src/consumer/topic.rs
```diff
@@ -95,45 +96,91 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
...
-                        if prev != addr {
-                            info!(
-                                "topic {} moved: previous = {:?}, new = {:?}",
-                                topic, prev, addr
-                            );
+                Err(ConnectionError::PulsarError(Some(err), text)) => {
+                    if matches!(
+                        err,
+                        proto::ServerError::ServiceNotReady | proto::ServerError::ConsumerBusy
+                    ) {
```
tisonkun: The `if` condition can be lifted into the match arm.

src/consumer/topic.rs
```diff
@@ -95,45 +96,91 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
...
+                        // Pulsar retryable error
+                        match operation_retry_options.max_retries {
+                            Some(max_retries) if current_retries < max_retries => {
```
tisonkun: Previously this branch matches `None` also?

src/consumer/topic.rs
```diff
@@ -95,45 +96,91 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
...
-                            Some(proto::ServerError::ServiceNotReady),
-                            text,
-                        )
-                        .into());
                     }
                 }
+                Err(ConnectionError::Io(e)) => {
+                    // The error is not retryable
+                    return Err(ConsumerError::Io(e).into());
+                }
```
tisonkun: Perhaps just fallthrough to `Err(e) => return Err(Error::Connection(e)),`?

src/consumer/topic.rs
```diff
@@ -95,45 +96,91 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
...
+                            current_retries += 1;
+                            client
+                                .executor
+                                .delay(operation_retry_options.retry_delay)
+                                .await;
+
+                            let addr = client.lookup_topic(&topic).await?;
+                            connection = client.manager.get_connection(&addr).await?;
+
+                            continue;
```
tisonkun: Why do we use different re-lookup logic here compare to the snippet above?

```rust
                            // we need to look up again the topic's address
                            let prev = addr;
                            addr = client.lookup_topic(&topic).await?;
                            if prev != addr {
                                info!(
                                    "topic {} moved: previous = {:?}, new = {:?}",
                                    topic, prev, addr
                                );
                            }

                            connection = client.manager.get_connection(&addr).await?;
                            continue;
```

tisonkun (commented):
Thanks for your contribution! Comments inline.

I'm looking into the new drop channel logic now.

src/consumer/engine.rs
```diff
@@ -571,35 +613,178 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
...
+    async fn reconnect(&mut self) -> Result<(), Error> {
+        debug!("reconnecting consumer for topic: {}", self.topic);
+        // Sender::send() method consumes the sender
+        // as the sender is hold by the ConsumerEngine, there is no way to call send method
+        // The lines below take the pointed sender and replace it by a new one bound to nothing
+        // but as the TopicProducer sender is recreate below, there is no worry
+        let (drop_signal, _) = oneshot::channel::<()>();
+        let old_signal = std::mem::replace(&mut self.drop_signal, drop_signal);
+        // This line ask for kill the previous errored consumer
+        let _ = old_signal.send(());
```
tisonkun: I'm thinking to use `Option<oneshot::Sender<()>>` to hold the sender and simply replace with None to trigger `drop`.

src/consumer/engine.rs
```diff
@@ -571,35 +613,178 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
...
+                            error!("subscribing({}) reached max retries", topic);
+                            Err(Err(Error::Connection(ConnectionError::Io(e))))
+                        }
+                    }
+                }
+            Err(ConnectionError::Io(e)) => {
+                // Non retryable IO Error
+                error!("reconnection got io error: {:?}", e);
+                return Err(Err(ConsumerError::Connection(ConnectionError::Io(e)).into()));
+            }
```
tisonkun: ditto I guess we can just fall through here.

tisonkun (commented):
Comments inline. Others LGTM.

I'm preparing a patch onto your branch so that you can directly make use of.

And here it is: https://github.com/CleverAkanoa/pulsar-rs/pull/2

src/consumer/topic.rs
CleverAkanoa: The None can be considered as the variant, we have an max_retries but it was exausted.

But if you prefer, we can split between `None` and `Some(max_retries) if current_retries >= max_retries)`

src/consumer/topic.rs
CleverAkanoa: Which one ?

tisonkun: It's included in https://github.com/CleverAkanoa/pulsar-rs/pull/2. Resolved.

src/consumer/topic.rs
tisonkun: I read the previous condition guard this block is:

```rust
                    if operation_retry_options.max_retries.is_none()
                        || operation_retry_options.max_retries.unwrap() > current_retries
```

but now:

```rust
                        match operation_retry_options.max_retries {
                            Some(max_retries) if current_retries < max_retries => {
```

tisonkun:
@CleverAkanoa Two conversations block this patch now:

* https://github.com/streamnative/pulsar-rs/pull/244#discussion_r1055129266
* https://github.com/streamnative/pulsar-rs/pull/244#discussion_r1054266529

After they get resolved, I think it's good to merge :)

src/consumer/topic.rs
FlorentinDUBOIS: @CleverAkanoa ping :D

FlorentinDUBOIS (approved):

CleverAkanoa:
Hello @tisonkun :)

Happy new year! 

Sorry, I was on holiday, thanks to all the improvements you have done on this PR. 

I'm doing the required modifications.

src/consumer/topic.rs
FlorentinDUBOIS: The condition is the same, but now max-retries is not at the right position of the condition

PR #252, opened 2023-01-27
Add the ability to `ack`, `nack` and `cumulative_ack` by message ID.

ChillFish8:
This PR adds the ability to `ack`, `nack` and `cumulative_ack` by message ID and topic name.

This is done because in some situations (like the one we have in Quickwit) we don't want to keep the original messages around or can't in some cases, instead using only the required metadata of the topic name and message-id data.

### DIFF
--- a/src/consumer/data.rs
+++ b/src/consumer/data.rs
@@ -16,8 +16,8 @@ pub enum EngineEvent<Exe: Executor> {
 }
 
 pub enum EngineMessage<Exe: Executor> {
-    Ack(MessageData, bool),
-    Nack(MessageData),
+    Ack(MessageIdData, bool),
+    Nack(MessageIdData),
     UnackedRedelivery,
     GetConnection(oneshot::Sender<Arc<Connection<Exe>>>),
 }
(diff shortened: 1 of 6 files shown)


FlorentinDUBOIS (approved):

FlorentinDUBOIS:
@tisonkun could you take a look at it as weel?

tisonkun (commented):
Generally looks good. Since `MessageData` is refactored to be unused here, could you @ChillFish8  remove the struct in this patch?

ChillFish8:
Yes no problem 👍

ChillFish8:
When checking this, we expose `MessageData` as a public API and use it to contain the additional batch metadata, removing it is a breaking change which I'm not sure we want to do.

tisonkun (approved):
@ChillFish8 Thanks for your investigation!

Then I'm OK to merge this patch as is. My opinion is that since `MessageIdData` contains `batchSize` field, the `MessageData` wrapper can be redundant. But we can investigate it later.

Merging...

PR #255, opened 2023-02-01
Make `Consumer::seek` future `Send`

guilload:
Currently, the future is not `Send`  because a non-`Send` type is held across the `.await` point. In this case, a mutable reference to `MultiTopicConsumer`. Cloning the config object before building the future solves the issue.

### DIFF
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -159,6 +159,7 @@ impl<T: DeserializeMessage, Exe: Executor> Consumer<T, Exe> {
             InnerConsumer::Multi(c) => {
                 c.seek(consumer_ids, message_id, timestamp).await?;
                 let topics = c.topics();
+                let config = c.config().clone();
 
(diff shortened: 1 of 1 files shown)


FlorentinDUBOIS (approved):

FlorentinDUBOIS:
Seems good to me! Thank a lot @guilload !

@CleverAkanoa and @tisonkun, we should discuss about surrounding producer and consumer configuration using an atomic reference counter. What do you think?

FlorentinDUBOIS (approved):

tisonkun (commented):
> surrounding producer and consumer configuration using an atomic reference counter

I agree such refactoring can ease development and may increase the efficiency of clones a bit - we don't have to optimize the access by using reference here.

tisonkun:
... while I found that `ProducerOptions` and `ConsumerConfig` are exported in public methods:

* TopicProducer#options
* TopicConsumer#config

so it can be a breaking change.

PR #256, opened 2023-02-02
Add support for on-the-fly AuthChallenge

klaatu01:
This PR Adds support for on-the-fly auth by handling the `AuthChallengeRequest` command.

Currently receiving `AuthChallengeRequest` logs an error and drops the command; this behaviour can lead to the client becoming disconnected. Instead, the client should authenticate with the Auth provider again and send an `AuthChallengeResponse` command containing the new `auth_data`, which will keep the connection alive.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1586,6 +1586,7 @@ mod tests {
     use std::{sync::Arc, time::Duration};
     use tokio::{net::TcpListener, sync::RwLock};
     use uuid::Uuid;
+    use crate::proto::{AuthData, CommandAuthResponse};
 
     #[tokio::test]
     #[cfg(feature = "tokio-runtime")]
@@ -1641,7 +1642,7 @@ mod tests {
         }
         async fn auth_data(&mut self) -> Result<Vec<u8>, AuthenticationError> {
             *self.count.write().await += 1;
-            Ok(Vec::new())
+            Ok("test_auth_data".as_bytes().to_vec())
         }
     }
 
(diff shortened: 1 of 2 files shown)


tisonkun:
@klaatu01 Thanks for your contribution! I'll review it by the next week.

At the first glance, since it supports a new feat, could you add a test to verify that the function works properly? In the CI env we prepare a Pulsar server:

https://github.com/streamnative/pulsar-rs/blob/56adb96a6bbf101af3540102faf27da30758a036/.github/workflows/rust.yml#L37

klaatu01:
Hey @tisonkun, thanks for taking the time to look over this.
I have added two tests. I decided not to do an integration test against the pulsar server as it was too difficult to get the AuthChallenge command to trigger. In the unit tests, we are mocking the TcpStream to send the commands to the connection and created a mock authentication method that counts the number of times it was hit. 

Let me know if this is okay. Have a nice weekend.

FlorentinDUBOIS (approved):

klaatu01:
@FlorentinDUBOIS failed due to rate limiting on the protoc install, can we get a rerun?

tisonkun (approved):
LGTM. Thanks for your contribution @klaatu01!

Although I have a concern that we hardcode the connector protocol version as 12 in:

```rust
        Message {
            command: proto::BaseCommand {
                r#type: CommandType::Connect as i32,
                connect: Some(proto::CommandConnect {
                    auth_method_name,
                    auth_data,
                    proxy_to_broker_url,
                    client_version: String::from("2.0.1-incubating"),
                    protocol_version: Some(12),
                    ..Default::default()
                }),
                ..Default::default()
            },
            payload: None,
        }
```

... while the first protocol version including auth challenge is v14 https://github.com/apache/pulsar/pull/3677. But it seems that the server doesn't verify it so this change can be correct. We may visit the hardcode protocol version and client version later.

tisonkun:
Merging...

klaatu01:
@tisonkun 
> LGTM. Thanks for your contribution @klaatu01!
> 
> Although I have a concern that we hardcode the connector protocol version as 12 in:
> 
> ```rust
>         Message {
>             command: proto::BaseCommand {
>                 r#type: CommandType::Connect as i32,
>                 connect: Some(proto::CommandConnect {
>                     auth_method_name,
>                     auth_data,
>                     proxy_to_broker_url,
>                     client_version: String::from("2.0.1-incubating"),
>                     protocol_version: Some(12),
>                     ..Default::default()
>                 }),
>                 ..Default::default()
>             },
>             payload: None,
>         }
> ```
> 
> ... while the first protocol version including auth challenge is v14 [apache/pulsar#3677](https://github.com/apache/pulsar/pull/3677). But it seems that the server doesn't verify it so this change can be correct. We may visit the hardcode protocol version and client version later.

Agreed.
It seems like most implementations return a constant for the version that the client currently supports. 
https://github.com/apache/pulsar-client-go/blob/cb983a7ebe1390b75832c58c15acb28796c35983/pulsar/internal/connection.go#L799-L806 

But yes, good to keep in mind.

Thanks for getting that merged!

PR #258, opened 2023-02-12
feat: transactions

### LINKED ISSUE #253: Transaction API support
It'd be nice to support [transactions API](https://pulsar.apache.org/docs/2.11.x/txn-use/)

morigs:
Resolves #253


zcourts:
Hi @morigs 
Is there a blocker preventing this moving forward? 

We're forced to use the Java client at the moment because tx support isn't widely available in Rust or C++ clients yet but it'll make things a lot better for us to merge that into our rust impl and drop the Java service which only exist because of this missing feature.

morigs:
@zcourts 
No blocker. Adding transactions support is in fact a big epic and requires a lot of effort.
I'm not that familiar with pulsar, so it's better for some of the contributors to decompose the task. In that case me and other contributors will be able to do separate isolated issues.
Unfortunately, I don't have enough time and expertise to do all this myself :(

PR #260, opened 2023-02-15
Add important documents notes related to delays and shedule message configs.

grv07:


tisonkun (approved):
Thanks for your contribution @grv07! Merging...

PR #266, opened 2023-03-18
Use atomic long to gen producer/consumer ids and simplify debugging

### LINKED ISSUE #265: Producer and Consumer IDs
The current producer and consumer ids are random unsigned longs, by default. This behavior diverges from some of the other Pulsar clients. The Java client has atomic longs that are incremented for each new producer/consumer. While it is certainly not a requirement to follow that convention, one interesting consequence of the current rust client's implementation is that the producer and consumer ids will not match when the unsigned long maps to a negative signed long because the broker logs will map to signed longs.

Here is a sample broker log with the producer id's negative number that did not match the rust client's logs:

> 2023-03-17T23:50:51,093-0500 [broker-topic-workers-OrderedExecutor-1-0] INFO  org.apache.pulsar.broker.service.ServerCnx - [/127.0.0.1:54796] Created new producer: Producer{topic=PersistentTopic{topic=persistent://public/default/test}, client=/127.0.0.1:54796, producerName=my producer, producerId=-2871063498580218646}

michaeljmarshall:
Fixes: #265 

The issue has all of the context. The general idea is to improve debugging given that Java is mapping an unsigned long to a signed long and then logging that value.

I am still learning rust, so if any of my usage is incorrect, please correct it!

### DIFF
--- a/src/consumer/topic.rs
+++ b/src/consumer/topic.rs
@@ -29,7 +29,7 @@ use crate::{
     BrokerAddress, DeserializeMessage, Error, Executor, Payload, Pulsar,
 };
 
-static consumer_id_generator: AtomicU64 = AtomicU64::new(0);
+static CONSUMER_ID_GENERATOR: AtomicU64 = AtomicU64::new(0);
 
 // this is entirely public for use in reader.rs
 pub struct TopicConsumer<T: DeserializeMessage, Exe: Executor> {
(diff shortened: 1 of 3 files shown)


src/consumer/topic.rs
```diff
@@ -60,7 +63,7 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
             options,
             dead_letter_policy,
         } = config.clone();
-        let consumer_id = consumer_id.unwrap_or_else(rand::random);
+        let consumer_id = consumer_id.unwrap_or_else(|| CONSUMER_ID_GENERATOR.fetch_add(1, Ordering::SeqCst));
```
michaeljmarshall: I wonder why we allow users to configure this id. We don't do that in the Java client. It's not inherently wrong, but it seems like it might leak the abstraction and could have unintended consequences.

tisonkun: I agree that we can change this manner and remove the consumer id from the config struct.

tisonkun (approved):
LGTM. Fixing clippy and pending to merge...

tisonkun:
Merging...

If you're going to remove the consumer id from the config object, welcome to open an issue so that we can determinate if it's a breaking change or just fix an unintentional leak.

michaeljmarshall:
Thanks for fixing and merging the PR @tisonkun!

PR #267, opened 2023-03-21
Added support for the client to authenticate itself to the server using TLS

### LINKED ISSUE #197: Support for TLS Authentication
In the current `pulsar crate`, it is mentioned as a feature: `TLS connection`, 
but when using `fn with_auth()` of the `Struct pulsar::producer::Producer::Builder`, it does not support `TLS Authentication`.

The current Authentication support:`pub struct Authentication { /// Authentication kid. Use "token" for JWT
    pub name: String,
    /// Authentication data
    pub data: Vec<u8>,
}`

coming from `connection.rs` of the crate

Is there a way to use a `TLS Authentication` method, because there is no clear way to do so as for now

afonsosribeiro:
closes streamnative#197

The client is now able to provide it's certificate and private key to the server.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -506,6 +506,25 @@ impl<Exe: Executor> PulsarBuilder<Exe> {
         self
     }
 
+    /// add a certificate and private key to authenticate the client in TLS connections
+    #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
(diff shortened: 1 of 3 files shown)


danielmulvad:
Hi StreamNative team, I'm looking to get this commit merged. Please take a look at my commit linked above, and let me know what I can do to push this forward.

Issue #268, opened 2023-03-26
Limit producer batch messages by size in bytes as well


AfonsoRibeiro:
To prevent the producer from sending batches that are bigger than allowed by the pulsar server, a limit in bytes should be possible to configure.
Or at least a soft threshold that when reached sends the batch (much easier to implement and it appears it is also done in the go client).

tisonkun:
This is partially resolved by https://github.com/streamnative/pulsar-rs/pull/269.

Keep open for a while since there are still several potential improvements can be made. See https://github.com/streamnative/pulsar-rs/pull/269#pullrequestreview-1382771811.

fatelei:
i want to try this issue

PR #269, opened 2023-03-26
Soft treshhold for the max batch size in bytes

AfonsoRibeiro:
If this threshold is reached before batch size the batch is considered full and sent.
A soft threshold means there is no performance penalty with this implementation

#268

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -138,6 +138,9 @@ pub struct ProducerOptions {
     pub schema: Option<Schema>,
     /// batch message size
     pub batch_size: Option<u32>,
+    /// batch size in bytes treshold (only relevant when batch_size active).
+    /// batch is sent when batch size in bytes is reached
+    pub batch_byte_size: Option<usize>,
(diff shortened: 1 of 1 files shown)


tisonkun (approved):
Thanks for your contribution!

I notice that the Java client handle `CommandConnected` event and use `maxMessageSize` also. Perhaps we can have a follow-up to handle it also.

Moreover, we always check full after queue the new message so it's a threshold than a guard, while we may check size before actually `push_back`.

tisonkun:
cc @FlorentinDUBOIS @BewareMyPower for peer reviewing.

@AfonsoRibeiro I'll appreciate it if you can add a function test.

FlorentinDUBOIS (approved):
Seems good to me

BewareMyPower (approved):

tisonkun:
Merging...

Potential improvements mentioned above can be follow-ups.

Thanks a lot for your contribution @AfonsoRibeiro!

PR #271, opened 2023-04-04
Fix ConsumerEngine::reconnect loop

### LINKED ISSUE #263: Consumer stucks in reconnections
Hello everyone.

Want to report a bug.

### Versions:
pulsar-rs = 4.2.0 (also reproduced on 5.1.0)
apache pulsar = 2.9.2

### Relates:
https://github.com/streamnative/pulsar-rs/issues/189

### Attachments:
- Cargo.toml: https://gist.github.com/myromeu/c4d5ea48b2b2c5b989fea36665ea2508#file-cargo-toml
- main.rs: https://gist.github.com/myromeu/c4d5ea48b2b2c5b989fea36665ea2508#file-main-rs
- 4.2.0.log: https://gist.github.com/myromeu/c4d5ea48b2b2c5b989fea36665ea2508#file-4-2-0-log
- 5.1.0.log: https://gist.github.com/myromeu/c4d5ea48b2b2c5b989fea36665ea2508#file-5-1-0-log

### To reproduce:
- Start consumer and producer on topic (persistent or not doesn't matter) with env RUST_LOG=pulsar=debug.
- Make unload namespace with command: `bin/pulsar-admin namespaces unload foo/bar`. But sometimes it reproduces in another unclear circumstances(:
- After this consumer stucks in infinitely in reconnections (look attached files for logs).

### Own findings:
In ConsumerEngine after registering sources it iterates them and sends to the engine. And in main loop inside `ConsumerEngine::engine` while looping it tries to reconnect if encountered some problems with underlying connection. And when it got `None` from messages source it tries to reconnect. But while it's reconnecting it still receives `None`'s inside registered source's task. After it reconnected it goes back to the main loop and receives collected `None`'s which was collected previously. So after received `None` it should reconnect again. And engine stucks inside this loop: Got None -> Go to reconnect -> Reconnection done -> Got `None` again -> ...

I had solved this with set flag self.reconnected and skip messages with None in the main loop. After skipping None's it resets the flag. But it can skip None message which comes right after reconnection and signals about real disconnection. So it cannot be solution.

Wonshtrum:
This patch is to fix #263.

The first commit contains the main fix, which should be sufficient to reconnect a consumer correctly. The two fixes are, as discussed in the issue :
- ignore the drop of the old message source
- send the flow message over the new connection

The following commits aim to clean up some related parts of the connection and reconnection code, to make it clearer and hopefully less error-prone.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1130,7 +1130,7 @@ impl<Exe: Executor> Connection<Exe> {
 impl<Exe: Executor> Drop for Connection<Exe> {
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
     fn drop(&mut self) {
-        trace!("dropping connection {} for {}", self.id, self.url);
+        debug!("dropping connection {} for {}", self.id, self.url);
         if let Some(shutdown) = self.sender.receiver_shutdown.take() {
             let _ = shutdown.send(());
         }
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -46,6 +46,9 @@ pub(crate) enum Register {
         consumer_id: u64,
         resolver: mpsc::UnboundedSender<Message>,
     },
+    RemoveConsumer {
+        consumer_id: u64,
+    },
     Ping {
         resolver: oneshot::Sender<()>,
     },
@@ -156,6 +159,9 @@ impl<S: Stream<Item = Result<Message, ConnectionError>>> Future for Receiver<S> 
                     resolver,
                 })) => {
                     self.consumers.insert(consumer_id, resolver);
+                }
+                Poll::Ready(Some(Register::RemoveConsumer { consumer_id })) => {
+                    self.consumers.remove(&consumer_id);
                 }
                 Poll::Ready(Some(Register::Ping { resolver })) => {
                     self.ping = Some(resolver);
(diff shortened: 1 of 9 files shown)


tisonkun:
Thanks for your contribution @Wonshtrum!

I can find only one commit in this patch, it seems not follow your description. Can you check if there is some mistake here?

Wonshtrum:
I was planning to push the other commits shortly afterward but was delayed by unexpected bugs.
The second commit focuses on `drop_signal` by refactoring it to make it simpler.
I intend to push a third commit merging duplicated code related to the connection (which made debugging the original issue much harder).

FlorentinDUBOIS (approved):
LGTM

KannarFr:
@tisonkun can you review it too? :)

tisonkun:
I'll review this patch tomorrow.

tisonkun (approved):
Thanks for your contributions @Wonshtrum!

The background descriptions and separate commits are very clean to review. I'm very happy to review your patch XD.

One comment here: perhaps we need an integration test for guarding this case. But it seems that such a test should run isolated and using pulsar-admin script to instruct changes. Thus, I don't block this PR for adding such a test.

tisonkun:
Also, I'm curious about your use case and see how pulsar-rs can satisfy your requirements. Looking forward to further communication - I'm `@tison` on the [Pulsar slack workspace](https://communityinviter.com/apps/apache-pulsar/apache-pulsar) or you can simply mailto:wander4096@gmail.com.

tisonkun:
Merging...

Maybe we can fire a 5.1.1 release for downstream testing. What do you think @FlorentinDUBOIS @KannarFr?

KannarFr:
Let's release this.

Wonshtrum:
I do not intend to push any more commits on this branch. You can release it. I hope I didn't break anything!

FlorentinDUBOIS:
@tisonkun, @Wonshtrum @KannarFr and I are working at Clever Cloud, we use Apache Pulsar as the main event bus of the company. It is mainly use between control planes micro-services and between control plane micro-services and data plane components.

tisonkun:
Released 5.1.1 - https://crates.io/crates/pulsar/5.1.1

PR #272, opened 2023-04-05
Add feature to compile protobuf library from source

liamkinne:
This is useful for cross compilation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ tokio = { version = "^1.21.2", features = ["macros", "rt-multi-thread"] }
 
 [build-dependencies]
 prost-build = "^0.11.1"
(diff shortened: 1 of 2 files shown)


tisonkun:
Thank you! Can we remove `setup-protoc` step with this patch?

tisonkun:
I'm merging this patch first.

You're welcome to figure out if we can remove `setup-protoc` step in the workflows with this patch.

tisonkun:
Thanks for your contribution @liamkinne!

liamkinne:
@tisonkun If its going to cause the build time to go up significantly, it may not be worth it. Will give it a go and report back.

tisonkun:
Thank you!

PR #278, opened 2023-05-25
Remove un-subscription when closing consumer

klaatu01:
Remove 'unsubscribe' from `consumer.close()`

https://github.com/streamnative/pulsar-rs/issues/279

### DIFF
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -254,7 +254,6 @@ impl<T: DeserializeMessage, Exe: Executor> Consumer<T, Exe> {
     }
 
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
-    /// Unsubscribe the topic then close the connection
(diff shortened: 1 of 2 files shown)


klaatu01:
@tisonkun I understand that this can be a breaking change for some. Do we want to consider making a new function?

tisonkun (approved):
Thank you! I check the behavior of Java client and agree with your change.

tisonkun:
> @tisonkun I understand that this can be a breaking change for some. Do we want to consider making a new function?

I remember that @DonghunLouisLee proposed https://github.com/streamnative/pulsar-rs/issues/224 for version control. I feel somehow this lib is still under development before calling it a formal semantic version - We call for 5.0 at a small configuration update https://github.com/streamnative/pulsar-rs/pull/240.

But in another way, big version number doesn't mean a bad thing (IIRC arrow-rs arrives v40+ already).

So my preference now is - release a 6.0.0 for this patch.

PR #281, opened 2023-06-02
ci: simplify protoc dep and update pulsar test broker versions

tisonkun:
This follows up #272. I test locally and it seems build from source doesn't take too much time.


tisonkun:
Merging...

PR #282, opened 2023-06-07
feat: use package version for connect client version

KannarFr:


src/message.rs
```diff
@@ -638,7 +638,6 @@ mod tests {
 
         {
             let connect = message.command.connect.as_ref().unwrap();
-            assert_eq!(connect.client_version, "2.0.1-incubating");
```
tisonkun, line 641: Let's update the assertion instead of remove it?

tisonkun: 1. follow the same rule of generation
2. doesn't malformed (i.e., the `env!("CARGO_PKG_VERSION")` part returns not empty or even in pattern `X.Y.Z`)

KannarFr: The assertion is based on the bytes array input of a message which we won't update for each version of pulsar-rs? WDYT?

Not sure to understand your second comment?

tisonkun: Make sense. I push one commit to update the bytes so that we at least sync once with the new change.

KannarFr: LGTM. Ty!

tisonkun (approved):
Merging...

Thanks for your contribution @KannarFr!

PR #286, opened 2023-07-17
fix: dead letter policy for batched messages

### LINKED ISSUE #174: Implement dead letter queue for unacked batched messages
Currently automatically sending unacked messages to dead letter queue only works with single messages, for batched messages the implementation appears to be postponed (there is a `TODO` message https://github.com/wyyerd/pulsar-rs/blob/master/src/consumer.rs#L1164). Pulsar-admin and our other producer implementation appears to be sending messages batched, so I had to implement this myself using a local counter. Having this coded in DB would be more optimal.

rliang:
Fixes `// TODO: Dead letter policy for batched messages` in `engine.rs`.

Probably not the most performant implementation right now since it collects all payloads into a single Vec instead of iterating over it, but it seems to get the job done for now.

Closes #174

### DIFF
--- a/src/consumer/engine.rs
+++ b/src/consumer/engine.rs
@@ -484,34 +484,31 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
             }
         };
 
-        match payload.metadata.num_messages_in_batch {
-            Some(_) => {
-                let it = BatchedMessageIterator::new(message.message_id, payload)?;
-                for (id, payload) in it {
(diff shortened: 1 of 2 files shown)


tisonkun:
Thanks for your contribution! I'll review this patch in this week.

BewareMyPower:
Could you resolve the conflicts?

rliang:
Done!

BewareMyPower (approved):

PR #288, opened 2023-07-26
build: allow to use rustls instead of native-tls 

FlorentinDUBOIS:
* This is used in an effort to remove all dependencies to openssl.
  Which could be interesting in embedded system or on environment
  which is difficult to know on which openssl version the software
  will run it and breaks deployments.
* It introduces two compiler feature flags which are `tokio-rustls-runtime`
  and `async-std-rustls-runtime` that have the same meaning as
  `tokio-runtime` and `async-std-runtime` except that they use rustls.
* There is a safe guard, if we enable both runtimes, this is the
  native-tls ones that are used to keep consistent with the current
  behaviour.

### DIFF
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -437,11 +437,11 @@ mod tests {
     };
     use log::LevelFilter;
     use regex::Regex;
-    #[cfg(feature = "tokio-runtime")]
+    #[cfg(any(feature = "tokio-runtime", feature = "tokio-rustls-runtime"))]
     use tokio::time::timeout;
 
     use super::*;
-    #[cfg(feature = "tokio-runtime")]
+    #[cfg(any(feature = "tokio-runtime", feature = "tokio-rustls-runtime"))]
     use crate::executor::TokioExecutor;
     use crate::{
         consumer::initial_position::InitialPosition, producer, proto, tests::TEST_LOGGER,
@@ -476,7 +476,7 @@ mod tests {
         tag: "multi_consumer",
     };
     #[tokio::test]
-    #[cfg(feature = "tokio-runtime")]
+    #[cfg(any(feature = "tokio-runtime", feature = "tokio-rustls-runtime"))]
     async fn multi_consumer() {
         let _result = log::set_logger(&MULTI_LOGGER);
         log::set_max_level(LevelFilter::Debug);
@@ -567,7 +567,7 @@ mod tests {
     }
 
     #[tokio::test]
-    #[cfg(feature = "tokio-runtime")]
+    #[cfg(any(feature = "tokio-runtime", feature = "tokio-rustls-runtime"))]
     async fn consumer_dropped_with_lingering_acks() {
         use rand::{distributions::Alphanumeric, Rng};
         let _result = log::set_logger(&TEST_LOGGER);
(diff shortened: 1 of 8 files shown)


FlorentinDUBOIS:
Hello @tisonkun :wave:, long time no see ! 
I made some changes to allow to use rustls instead of native-tls using compiler feature flags.
Could you give me your thought of this pull request ?

To give you more context about this pull request, at Clever Cloud, we are doing cross-compiling stuffs and some times build statically binaries and it will helps us, if we could remove the dependency to openssl to do so.

FlorentinDUBOIS:
By the way, I have updated all dependencies !

FlorentinDUBOIS:
Hello @tisonkun and @CleverAkanoa, I have deployed this patch in production and it works pretty well.

CleverAkanoa (approved):
Looks good :)

tisonkun:
Thank you! I'll review this patch in days.

src/consumer/mod.rs
```diff
@@ -437,11 +437,11 @@ mod tests {
     };
     use log::LevelFilter;
     use regex::Regex;
-    #[cfg(feature = "tokio-runtime")]
+    #[cfg(any(feature = "tokio-runtime", feature = "tokio-rustls-runtime"))]
```
tisonkun, line 440: Where do we define `tokio-rustls-runtime` feature?

src/error.rs
```diff
@@ -113,13 +124,36 @@ impl From<io::Error> for ConnectionError {
...
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
     fn from(err: native_tls::Error) -> Self {
         ConnectionError::Tls(err)
     }
 }
 
+#[cfg(all(
+    any(
+        feature = "tokio-rustls-runtime",
+        feature = "async-std--rustls-runtime"
```
tisonkun:
```suggestion
        feature = "async-std-rustls-runtime"
```

typo?

src/consumer/mod.rs
tisonkun: I saw the new definitions in `Cargo.toml` now.

tisonkun (changes requested):
Thanks for your contribution @FlorentinDUBOIS!

The direction looks good to me. One minor comment inline.

Also, I'm thinking of introduce a compile error guard to assert users not enable rustls and native-tls at the same time. This can simplify the cfg guard and give users a clear expectation. Existing configs won't be affected also.

See https://doc.rust-lang.org/std/macro.compile_error.html for reference.

But I can accept if we do this refactor later after this PR merged and before the next release out.

FlorentinDUBOIS:
I am glad that you are ok with those modification, I will do the modification above about the hyphen and I will add the compile time error as well.
By the way, once this will be merged I will rebase the pull request #267  to integrate those changes.

src/consumer/mod.rs
FlorentinDUBOIS: Yes, exactly :)

FlorentinDUBOIS:
Hey @tisonkun, I did the modification, could you check that everything is ok for you ?

tisonkun (approved):
Thank you!

FlorentinDUBOIS:
As @CleverAkanoa and @tisonkun approve this pull request, I merge it.

PR #289, opened 2023-08-06
build: disable oldtime feature of chrono

paolobarbolini:
`chrono` enables the `oldtime` feature for backwards compatibility but this project doesn't make use of it. Disabling it avoids the dependency on `time` v0.1


tisonkun (approved):
Thank you!

tisonkun:
Merging...

PR #291, opened 2023-08-11
chore: Implement Sync for Consumer

cirias:
Hi! This request adds the auto implementation of `Sync` for `Consumer`.

Currently, `Sync` is not implemented for `Consumer`, which causes error when trying to move the consumer to a backgound task. See the code below as an example.

```
async main() {
    let consumer: Consumer<TestData, _> = pulsar
        .consumer()
        .build()
        .await?;

    let mut app = App { consumer };

    tokio::task::spawn(app.run()).await.unwrap();
}

struct App {
    consumer: Consumer<TestData, TokioExecutor>,
}

impl App {
    async fn run(&mut self) {
        while let Some(_msg) = self.consumer.try_next().await.unwrap() {
            self.dosomething().await;
        }
    }

    async fn dosomething(&self) {}
}
```

---
- Updated the example.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -871,10 +871,11 @@ impl<Exe: Executor> Connection<Exe> {
         match auth {
             Some(m_auth) => {
                 let mut auth_guard = m_auth.lock().await;
-                Ok(Some(Authentication {
-                    name: auth_guard.auth_method_name(),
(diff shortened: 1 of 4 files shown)


tisonkun (approved):
Thank you!

PR #294, opened 2023-08-29
forward ordering key

stellit:
forward ordering key while preparing the message


BewareMyPower (commented):
Could you add a test to protect this change?

IIRC, without this patch, the message received from the consumer won't have the ordering key that was set by the producer side.

BewareMyPower:
Close it due to long inactivity time

PR #297, opened 2023-09-20
refactor: Remove mut from `send` for producer

cirias:
The pulsar client in other languages support `send_async` which increases the throughput a lot because it doesn't wait the response of each `send` of message.
I was trying to add a similar thing in the rust pulsar client. Instead of adding a new method, the user can just `spawn` the `send` future, so it becomes async automatically. But one thing that prevents user doing that is, the `send` method takes `&mut self`. Otherwise, the user can wrap the producer in `Arc` and clone it for each `send`.
In the client, for each connection, it has a queue contains all the messages/commands to be sent, and the `send` method just pushes a command to that queue. So it feels like shouldn't be hard to remove the `mut`.
But then I found the only thing that needs the `mut` is when connection failed, it reconnects and then updates the `connection` field of the producer. That means we have to wrap the connection with a lock, but I wonder if it is too heavy.

Other thing is by having `mut` it naturally prevents multiple `reconnect` running in parallel. To remove `mut` we need consider that as well, which I haven't addressed in the PR yet, but want get some feedback first.

So do you thing this is the right approach to achieve `send_async`?

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -4,7 +4,7 @@ use std::{
     io::Write,
     pin::Pin,
     sync::{
-        atomic::{AtomicU64, Ordering},
+        atomic::{AtomicU64, AtomicUsize, Ordering},
         Arc,
     },
     time::{Duration, SystemTime, UNIX_EPOCH},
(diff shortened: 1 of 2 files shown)


cirias:
Replaced by https://github.com/streamnative/pulsar-rs/pull/298

PR #298, opened 2023-09-26
feat: Change send to not block on waiting receipt

cirias:
This is a follow up of https://github.com/streamnative/pulsar-rs/pull/297.

I figured it's probably better to reuse the `SendFuture` for async send, since the batch send already delays the resolve of `SendFuture`.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -538,7 +538,7 @@ impl<Exe: Executor> TopicProducer<Exe> {
                 };
 
                 trace!("sending a batched message of size {}", message_count);
-                let send_receipt = self.send_compress(message).await.map_err(Arc::new);
+                let send_receipt = self.send_compress(message).await?.await.map_err(Arc::new);
                 for resolver in receipts {
                     let _ = resolver.send(
                         send_receipt
@@ -557,8 +557,13 @@ impl<Exe: Executor> TopicProducer<Exe> {
         let (tx, rx) = oneshot::channel();
         match self.batch.as_ref() {
             None => {
-                let receipt = self.send_compress(message).await?;
-                let _ = tx.send(Ok(receipt));
+                let fut = self.send_compress(message).await?;
+                self.client
+                    .executor
+                    .spawn(Box::pin(async move {
+                        let _ = tx.send(fut.await);
+                    }))
+                    .map_err(|_| Error::Executor)?;
                 Ok(SendFuture(rx))
             }
             Some(batch) => {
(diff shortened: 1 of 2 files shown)


src/producer.rs
```diff
@@ -557,8 +557,13 @@ impl<Exe: Executor> TopicProducer<Exe> {
         let (tx, rx) = oneshot::channel();
         match self.batch.as_ref() {
             None => {
-                let receipt = self.send_compress(message).await?;
-                let _ = tx.send(Ok(receipt));
+                let fut = self.send_compress(message).await?;
+                self.client
+                    .executor
+                    .spawn(Box::pin(async move {
+                        let _ = tx.send(fut.await);
```
mattisonchao, line 564: We put the request to a multi-thread executor. Is it possible to disorder the message?

mattisonchao: cc @cirias

cirias: @mattisonchao Thank you for reviewing this!

The order of the messages should remain the same as the order of the `send`. Since each message is pushed to one channel underneath. https://github.com/cirias/pulsar-rs/blob/feature/non-blocking-send/src/connection.rs#L664

`send_compress` still runs synchronously, and the pushing to channel action of the link above happens inside it.

mattisonchao (approved):

mattisonchao:
Hi, @cirias 
Btw, do you have any reports for performance improvement?

tisonkun (approved):
Thank you! LGTM.

cirias:
> Hi, @cirias Btw, do you have any reports for performance improvement?

I've done some not very scientific test. My goal is to send the messages in 40 milliseconds for each, then the messages per second would be 1000ms / 40ms = 25mps. Without the change, it can only reach 4mps. While after applying the change, it reaches 25mps easily. I'll attach the code I used for testing below. Also notice that the ping from the test machine to the pulsar cluster is about 200ms, which probably reduces the mps significantly when each send is blocked for the receipt.


```
async fn produce(
    mut producer: pulsar::Producer<pulsar::TokioExecutor>,
) -> Result<(), BoxedError> {
    let mut count = 0;
    let mut itv = tokio::time::interval(std::time::Duration::from_millis(40));
    let start_time = std::time::Instant::now();
    loop {
        itv.tick().await;
        let message = mock_message();
        producer.send(message).await?;
        count += 1;
        let mps = count as f32 / start_time.elapsed().as_secs_f32();
        tracing::info!("messages per second: {mps:.2}");
    }
}
```

Edit: I also have a consumer running on the same test machine. The message receiving rate matches to the producing rate.

tisonkun:
@cirias do you expect a new release for this patch? If so, I can prepare a 6.1.0 version for it.

cirias:
> @cirias do you expect a new release for this patch? If so, I can prepare a 6.1.0 version for it.

Yes! I'm currently referencing the commit hash directly in Cargo.toml. But it would be much better to have a version. Thank you!

tisonkun:
@cirias Thanks for your information!

I'm going to include https://github.com/streamnative/pulsar-rs/pull/299 also in 6.1.0 and will appreciate it if you can take a look at its description and changes to check the manner "fix" align with your use case - that is, do not auto redirect broker URL to tls scheme if non-tls broker URL set.

If it's OK, I'll merge it and shceulde the release tomorrow :D

cirias:
@tisonkun Thanks for the information!

The PR looks good to me. It shouldn't cause problems in my use case. I left some comments but it's totally fine for you to merge and release. Good luck!

tisonkun:
v6.1.0 is avaiable now - https://crates.io/crates/pulsar/6.1.0

PR #299, opened 2023-10-13
fix: do not auto redirect to tls scheme

WarmSnowy:
This patch fix the problem that the client will
auto redirect to tls scheme even setting non-tls broker url which may confuse the user.

### DIFF
--- a/src/service_discovery.rs
+++ b/src/service_discovery.rs
@@ -117,7 +117,7 @@ impl<Exe: Executor> ServiceDiscovery<Exe> {
             is_authoritative = authoritative;
 
             // Use broker url with the same schema of manager.url
-            let (connection_url, broker_url) = match self.manager.url.scheme() {
+            let (connection_url, broker_url) = match base_url.scheme() {
                 "pulsar+ssl" => {
                     if let Some(u) = &broker_url_tls {
                         (
--- a/src/service_discovery.rs
+++ b/src/service_discovery.rs
@@ -116,13 +116,32 @@ impl<Exe: Executor> ServiceDiscovery<Exe> {
             } = convert_lookup_response(&response)?;
             is_authoritative = authoritative;
 
-            // use the TLS connection if available
-            let connection_url = if let Some(u) = &broker_url_tls {
-                u.clone()
-            } else if let Some(u) = &broker_url {
-                u.clone()
-            } else {
-                return Err(ServiceDiscoveryError::NotFound);
+            // Use broker url with the same schema of manager.url
+            let (connection_url, broker_url) = match self.manager.url.scheme() {
+                "pulsar+ssl" => {
+                    if let Some(u) = &broker_url_tls {
+                        (
+                            u.clone(),
+                            format!("{}:{}", u.host_str().unwrap(), u.port().unwrap_or(6651)),
+                        )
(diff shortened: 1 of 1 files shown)


WarmSnowy:
@tisonkun Hi, is there a plan to merge this patch?

src/service_discovery.rs
```diff
@@ -116,13 +116,32 @@ impl<Exe: Executor> ServiceDiscovery<Exe> {
...
 
-            // use the TLS connection if available
-            let connection_url = if let Some(u) = &broker_url_tls {
-                u.clone()
-            } else if let Some(u) = &broker_url {
-                u.clone()
-            } else {
-                return Err(ServiceDiscoveryError::NotFound);
+            // Use broker url with the same schema of manager.url
+            let (connection_url, broker_url) = match self.manager.url.scheme() {
```
tisonkun:
```suggestion
            let (connection_url, broker_url) = match base_url.scheme() {
```

tisonkun (approved):
Thanks for your contribution! This patch is good to go.

tisonkun:
Although I wonder if you can add a function test to prevent the manner from being changed later, it's a bit difficult with our current codebase and test framework. I'll appreciate it a lot if you can achieve it while I don't make it a merge blocker.

Pending to merging once CI gives green.

src/service_discovery.rs
```diff
@@ -116,13 +116,32 @@ impl<Exe: Executor> ServiceDiscovery<Exe> {
             } = convert_lookup_response(&response)?;
             is_authoritative = authoritative;
 
-            // use the TLS connection if available
-            let connection_url = if let Some(u) = &broker_url_tls {
```
cirias, line 120: Why is `broker_url_tls` not None when the server doesn't support tls? Is it possible to be a misconfiguration?

src/service_discovery.rs
```diff
@@ -116,13 +116,32 @@ impl<Exe: Executor> ServiceDiscovery<Exe> {
...
+                }
+                "pulsar" => {
+                    if let Some(u) = &broker_url {
+                        (
+                            u.clone(),
+                            format!("{}:{}", u.host_str().unwrap(), u.port().unwrap_or(6650)),
+                        )
+                    } else {
+                        return Err(ServiceDiscoveryError::NotFound);
+                    }
```
cirias: nit: Not really a problem, but feel it's better to deduplicate the code here. Consider letting `match` return `(broker_url_maybe_tls, broker_port)`, then do `let connection_url = if let Some(..) = &broker_url_maybe_tls {` after.

src/service_discovery.rs
WarmSnowy: The `broker_url_tls` on here is parsed from server which the old client logic used to connect to server even if the configuration is non-tls url. It makes some trouble such as certificate verification problem even user not want to use tls.

cirias: I was thinking if it's possible there is a server configuration causes the server to send the tls url even it doesn't handle tls. But that is not related to your change. All good, thank you!

cirias (commented):
lgtm

tisonkun:
Thanks for your contribution and review!

PR #300, opened 2023-10-25
Add example to consume AVRO schema message produced by Java

zymap:
---

### Motivation

Add an example to consume AVRO schema messages produced by Java producer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ tracing = { version = "^0.1.37", optional = true }
 async-trait = "^0.1.72"
 data-url = { version = "^0.3.0", optional = true }
 uuid = { version = "^1.4.1", features = ["v4", "fast-rng"] }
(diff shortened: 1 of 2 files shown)


tisonkun (approved):
LGTM. Since we don't abstract schema internally, it can take a large effort to add full schema support >_<

It would be better you can add a doc (file or comment) to describe how to run this example.

Cargo.toml
```diff
@@ -54,6 +54,7 @@ tracing = { version = "^0.1.37", optional = true }
 async-trait = "^0.1.72"
 data-url = { version = "^0.3.0", optional = true }
 uuid = { version = "^1.4.1", features = ["v4", "fast-rng"] }
+apache-avro = "0.16.0"
```
tisonkun, line 57: Move to `dev-dependencies` as we don't support it in lib yet.

PR #305, opened 2024-01-18
fix: end client rx upon disconnection error

rliang:
Currently, consumers can enter an infinite error logging loop when Pulsar goes down (even if it goes back up).

The cause seems to be that, when registering an event source, the send loop never exits, even if the event sink is disconnected. This patch fixes it.


klaatu01:
@tisonkun I can vouch for this issue: 
Screenshot 2024-03-05 at 11 02 50

tisonkun:
I don't have the permission to handle this repo traffic now, sorry.

freeznet (approved):

freeznet:
This PR lgtm, and i will merge it once the checks all passes.

@tisonkun just wanna to check with you if you are willing to continue work on this repo? if yes, sijie could grant the permissions to you so you could continue to work on this repo when you have time. thanks.

PR #311, opened 2024-05-10
feat: Add back blocking send method based on non-blocking version

cirias:
- Rename all non blocking send methods to be suffixed with `_non_blocking`. Per request from https://github.com/streamnative/pulsar-rs/pull/310.

- Fix missing image of pulsar in CI job.

### DIFF
--- a/examples/round_trip.rs
+++ b/examples/round_trip.rs
@@ -59,6 +59,8 @@ async fn main() -> Result<(), pulsar::Error> {
                     data: "data".to_string(),
                 })
                 .await
+                .unwrap()
+                .await
                 .unwrap();
             counter += 1;
             if counter % 1000 == 0 {
--- a/examples/round_trip.rs
+++ b/examples/round_trip.rs
@@ -59,8 +59,6 @@ async fn main() -> Result<(), pulsar::Error> {
                     data: "data".to_string(),
                 })
                 .await
-                .unwrap()
-                .await
                 .unwrap();
             counter += 1;
             if counter % 1000 == 0 {
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,10 +27,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        pulsar-version: [ 2.10.4.3, 2.11.1.2, 3.0.0.1 ]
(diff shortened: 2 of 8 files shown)


cirias:
@mattisonchao @tisonkun

examples/round_trip.rs
```diff
@@ -59,8 +59,6 @@ async fn main() -> Result<(), pulsar::Error> {
                     data: "data".to_string(),
                 })
                 .await
-                .unwrap()
```
mattisonchao: It looks like we are breaking the existing public API. isn't it?

cirias: It is. But for a good intention. In version 5.1.1, it [returns](https://github.com/streamnative/pulsar-rs/blob/v5.1.1/src/producer.rs#L552) a future of receipt when the receipt is already received from the server. It causes misunderstanding of that the client are still waiting for receipt while the client already waited for it. Even the [doc](https://github.com/streamnative/pulsar-rs/blob/v5.1.1/src/producer.rs#L318) was misleading too.

I made it to not return the future, but just the receipt, so it's clear that the receipt is awaited.

cirias:
Updated the commit message to make it clear that this is a breaking change.

examples/round_trip.rs
mattisonchao: Hi @cirias 

I am sorry. We should do some nasty things to avoid breaking the public API. Consider that users already have code written using the old approach, But the new version changes it directly. It's kinda pain for existing users.

We can keep the old approach, mark it deprecated, and remove it in the future. Also, we can introduce a brand new method to build an elegant API. 

I would like to know your thoughts. Thanks! :)

cirias: Make sense. I updated the methods to be backward compatible, and marked them deprecated.

mattisonchao (approved):
Thanks for your contribution.

mattisonchao:
HI, @cirias 
Would you like to continue to fixing the unbending channel issue for non-blocking send?

cirias:
> HI, @cirias Would you like to continue to fixing the unbending channel issue for non-blocking send?

@mattisonchao I would love to. I'm busy with my job recently, but will work on an PR once I have some time.

PR #320, opened 2024-07-30
chore: update dependencies and use send_non_blocking

FlorentinDUBOIS:
* If you are using rustls instead of native tls version of this crate (through `tokio-rustls-runtime` or `async-std-rustls-runtime`), you HAVE TO call the [`CryptoProvider::install_default`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default) method once in your binary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,63 +16,63 @@ description = "Rust client for Apache Pulsar"
 keywords = ["pulsar", "api", "client"]
 
 [dependencies]
-async-channel = "2"
-bytes = "^1.4.0"
-crc = "^3.0.1"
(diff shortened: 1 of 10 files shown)


CleverAkanoa:
There is a warning about the deprecation of the `send`  in doc tests
```
warning: use of deprecated method `pulsar::Producer::<Exe>::send`: instead use send_non_blocking
  --> src/lib.rs:78:14
   |
48 |             .send(TestData {
```
Same for:
- src/producer.rs:166:23
- src/producer.rs:167:23
- src/producer.rs:402:19
- src/producer.rs:403:19

CleverAkanoa (changes requested):
besides the doc test comment, seems good

PR #327, opened 2024-12-18
Fixes around service discovery, engine and consumer.

FlorentinDUBOIS:
* The first fix is around service discovery to not create lookup topic request amplification when we are able to determine if the topic is already a partition.
* The second fix is all about lookup request as well as there are performed systematically even though we aren't running a consumer with topic discovery based on a regex.
* The third fix is a tricky one that hit us in production at Clever Cloud, where our consumers have an issue with batched messages that create an underflow on the `remaining_messages` variable. The symptom was that consumer stop receiving messages, because they do not send any flow control message due to the invalid remaining message number. We also fix a bug that create a reconnect loop, because the driver consider that the newly created connection was already disconnected due to same origin.

### DIFF
--- a/src/consumer/engine.rs
+++ b/src/consumer/engine.rs
@@ -43,7 +43,7 @@ pub struct ConsumerEngine<Exe: Executor> {
     event_rx: mpsc::UnboundedReceiver<EngineEvent<Exe>>,
     event_tx: UnboundedSender<EngineEvent<Exe>>,
     batch_size: u32,
-    remaining_messages: u32,
+    remaining_messages: i64,
     unacked_message_redelivery_delay: Option<Duration>,
     unacked_messages: HashMap<MessageIdData, Instant>,
     dead_letter_policy: Option<DeadLetterPolicy>,
@@ -83,7 +83,7 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
             event_rx,
             event_tx,
             batch_size,
-            remaining_messages: batch_size,
+            remaining_messages: batch_size as i64,
             unacked_message_redelivery_delay,
             unacked_messages: HashMap::new(),
             dead_letter_policy,
@@ -108,10 +108,12 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
                     }
                 }
             }
+
             let send_end_res = event_tx.send(mapper(None)).await;
             if let Err(err) = send_end_res {
                 log::error!("Error sending end event to channel - {err}");
             }
+
             log::warn!("rx terminated");
         }))
     }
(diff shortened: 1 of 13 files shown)


FlorentinDUBOIS:
The dashboard that I use to check if the fix works :D We are able to treat batch messages!

image

src/consumer/engine.rs
```diff
@@ -43,7 +43,7 @@ pub struct ConsumerEngine<Exe: Executor> {
     event_rx: mpsc::UnboundedReceiver<EngineEvent<Exe>>,
     event_tx: UnboundedSender<EngineEvent<Exe>>,
     batch_size: u32,
-    remaining_messages: u32,
+    remaining_messages: i64,
```
CleverAkanoa, line 46: negative amount of messages are possible ?

src/consumer/engine.rs
```diff
@@ -146,37 +148,59 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
                     })?;
             }
 
-            if self.remaining_messages < self.batch_size.div_ceil(2) {
+            // debug!("self.remaining_messages ({}) < self.batch_size.div_ceil(2) ({}) = {}",
```
CleverAkanoa: are comments necessary?

src/consumer/engine.rs
```diff
@@ -146,37 +148,59 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
...
+            // self.remaining_messages, self.batch_size.div_ceil(2), self.remaining_messages <
+            // self.batch_size.div_ceil(2));
+            if self.remaining_messages < (self.batch_size.div_ceil(2) as i64) {
                 match self
                     .connection
                     .sender()
-                    .send_flow(self.id, self.batch_size - self.remaining_messages)
+                    .send_flow(
+                        self.id,
+                        (self.batch_size as i64 - self.remaining_messages) as u32,
```
CleverAkanoa, line 179: I don't know your attended behavior but if you cast a negative number to u32 this gives a big number

```
let a = 12_i64;
let b = 23_i64;
dbg!((a - b) as u32); // 4294967285
```

src/service_discovery.rs
```diff
@@ -176,9 +176,12 @@ impl<Exe: Executor> ServiceDiscovery<Exe> {
         &self,
         topic: S,
     ) -> Result<u32, ServiceDiscoveryError> {
-        let mut connection = self.manager.get_base_connection().await?;
         let topic = topic.into();
+        if topic.contains("-partition-") {
```
CleverAkanoa, line 191: maybe a comment to explain where does come from this magic string

CleverAkanoa (changes requested):
Some questions regarding hazardous casting

src/consumer/engine.rs
FlorentinDUBOIS: In the casual workflow, we use the `batch_size` as a maximum number that we could send using a send flow command message to ask the broker to send us messages, which is why we have a subtraction below (`batch_size` - `remaining_messages`).

In the special case of batch messages (which is defined by clients), the number of messages could be greater than the given batch size and the remaining messages goes negative, which is why we use an `i64` as we want to keep track of the number of negative messages as the next send flow will be the batch_size - minus remaining messages which allow us to retrieve the casual workflow of flow command messages. 

Here is the example of it works for a batch_size at 1000 and the error case: 

```
Message (1) -> (batch_size = 1000, remaing_messages = 999, no flow message trigger)
... 499 messages later ...
Message (1) -> (batch_size = 1000, remaing_messages = 499, flow message trigger, ask broker to send => batch_size - remaining_messages = 501)
... 200 messages later ...
BatchMessage (1024) -> (batch_size = 1000, remaining_messages = 4294967096, no flow message trigger) [underflow on remaining messages - without the patch]
BatchMessage (1024) -> (batch_size = 1000, remaining_messages = -1124, flow message trigger, ask broker to send =>  batch_size - remaining_messages = 2124) [no underflow on remaining messages - with the patch]
```

src/consumer/engine.rs
FlorentinDUBOIS: the batch size is always positive as it is a u32 and the remaining messages could be negative, but in that case it will output a positive number that fit in u32 as batch size is the maximum value that we want to converge.

FlorentinDUBOIS: Besides, if we ask of u32::MAX to the broker, it is not really an issue as well :P, it is better than receiving nothing

src/service_discovery.rs
FlorentinDUBOIS: yes, of course

CleverAkanoa (approved):
Better :)

KannarFr:
Who should review this? (and merge)

FlorentinDUBOIS:
Akanoa and I could do both, but I was waiting for people who works with at streamnative to review as well.

freeznet (approved):

PR #329, opened 2025-02-18
fix(retry): fix regression on the retry logic

mattisonchao:
### Motivation

The regression was introduced by https://github.com/streamnative/pulsar-rs/pull/244, which makes the undefined retry operation policy lose the semantic of infinity retry.


src/retry_op.rs
```diff
@@ -54,26 +54,23 @@ pub async fn handle_retry_error<Exe: Executor>(
...
-            error!(
-                "{operation_name}({topic}) answered {kind}{text}, retrying request after {:?} (max_retries = {max_retries})",
-                operation_retry_options.retry_delay
-            );
-            client
-                .executor
-                .delay(operation_retry_options.retry_delay)
-                .await;
+    if operation_retry_options.allow_retry(current_retries) {
+        let max_retries= operation_retry_options.max_retries.unwrap_or(0);
```
BewareMyPower:
```suggestion
        let max_retries = operation_retry_options.max_retries.unwrap_or(0);
```

Please run `cargo +nightly fmt --all --check` locally to make sure the format check pass

BewareMyPower: ~~Though I also found some existing code might be affected.~~ That's because my local code base has already applied another patch

src/producer.rs
```diff
@@ -1178,7 +1178,7 @@ impl<'a, T, Exe: Executor> MessageBuilder<'a, T, Exe> {
     }
 }
 
-impl<'a, T: SerializeMessage + Sized, Exe: Executor> MessageBuilder<'a, T, Exe> {
+impl<T: SerializeMessage + Sized, Exe: Executor> MessageBuilder<'_, T, Exe> {
```
BewareMyPower: This change seems not related to the fix. It's better to revert it.

src/retry_op.rs
```diff
@@ -54,26 +54,23 @@ pub async fn handle_retry_error<Exe: Executor>(
...
-        Some(max_retries) if current_retries < max_retries => {
-            error!(
-                "{operation_name}({topic}) answered {kind}{text}, retrying request after {:?} (max_retries = {max_retries})",
-                operation_retry_options.retry_delay
-            );
-            client
-                .executor
-                .delay(operation_retry_options.retry_delay)
-                .await;
+    if operation_retry_options.allow_retry(current_retries) {
```
BewareMyPower: Just my two cents: it would be better if short logic is put in the early return pattern, i.e.

```rust
if !ok {
    error!(msg);
    return Err(err.into());
}
// 11 lines
```

will be more clear than

```rust
if ok {
    // 10 lines
    return ok
}
error!(msg);
Err(err.into())
```

src/retry_op.rs
```diff
@@ -54,26 +54,23 @@ pub async fn handle_retry_error<Exe: Executor>(
...
+        let max_retries= operation_retry_options.max_retries.unwrap_or(0);
+        error!(
+            "{operation_name}({topic}) answered {kind}{text}, retrying request after {:?} (max_retries = {max_retries})",
```
BewareMyPower: `max_retries` here is only for logging. Calling `unwrap_or(0)` might be misleading that it does not differ 0 and `none`, you can print the option like

```rust
error!("max_retries = {:?}", operation_retry_options.max_retries);
```

src/retry_op.rs
mattisonchao: applied.

src/retry_op.rs
mattisonchao: applied.

src/retry_op.rs
mattisonchao: applied.

src/producer.rs
mattisonchao: Yep. It's reported by lint. let me revert it and use another PR to fix it then.

mattisonchao: https://github.com/streamnative/pulsar-rs/pull/330

BewareMyPower (approved):

PR #350, opened 2025-08-27
test(producer): add unit tests for producer.rs

ChrisWrenDev:
I've added unit tests to producer.rs to validate the producer’s local behavior ahead of structural refactors in upcoming PRs to reduce future sizes downstream. They are intentionally broker-free to keep the feedback loop tight and deterministic.

**Tests Added:**
send_future_errors_when_sender_dropped: 
- Ensures `SendFuture` returns `Error::Producer(ProducerError::Custom(..))` when the sender side is dropped (simulating an unexpected disconnect).

message_converts_into_producer_message:
- Confirms `From<Message>` for `ProducerMessage` copies user-facing fields and leaves producer-managed fields unset.

batch_fills_on_length_threshold_and_drains: 
- Length-based fullness at `Batch::new(3, None);` verifies drain resets state and that `BatchedMessage` metadata is populated correctly.

batch_fills_on_size_threshold_even_if_length_not_reached:
- Size-based fullness at `Batch::new(10, Some(10));` drain resets size accounting.

batch_is_full_when_size_exactly_matches_threshold:
- Confirms “== threshold” counts as full.

batch_mapping_when_optional_fields_absent:
- Optional fields remain unset in metadata; payload_size equals actual payload length.

batch_storage_tracks_and_resets_size:
- Direct test of `BatchStorage` internal size tracking and reset-on-drain behavior.

**Limitations / Future work:**
The remaining behavior I wanted to test is coupled to a real Pulsar connection. Without either (a) a running broker or (b) a mockable “transport” seam that pretends to be the broker, I'm unable to drive, observe, or deterministically fail the network paths.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -1205,3 +1205,270 @@ impl<T: SerializeMessage + Sized, Exe: Executor> MessageBuilder<'_, T, Exe> {
         producer.send_raw(producer_message).await
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::executor::block_on;
+
+    fn pm_with(
+        payload: Vec<u8>,
+        props: &[(&str, &str)],
+        partition_key: Option<&str>,
+        ordering_key: Option<&[u8]>,
+        event_time: Option<u64>,
+    ) -> ProducerMessage {
+        let mut properties = HashMap::new();
+        for (k, v) in props {
+            properties.insert((*k).to_string(), (*v).to_string());
+        }
+        ProducerMessage {
+            payload,
+            properties,
+            partition_key: partition_key.map(|s| s.to_string()),
+            ordering_key: ordering_key.map(|b| b.to_vec()),
+            event_time,
+            ..Default::default()
+        }
+    }
+
+    #[test]
+    fn send_future_errors_when_sender_dropped() {
+        let (tx, rx) = futures::channel::oneshot::channel::<Result<CommandSendReceipt, Error>>();
+        // Drop the sender immediately to simulate an unexpected disconnect:
+        drop(tx);
+
+        let fut = SendFuture(rx);
+        let err = block_on(fut).expect_err("expected an error when sender is dropped");
+
+        // It should be mapped to a ProducerError::Custom inside Error::Producer
+        match err {
+            Error::Producer(ProducerError::Custom(msg)) => {
+                assert!(
+                    msg.contains("unexpectedly disconnected"),
+                    "unexpected error message: {msg}"
+                );
+            }
+            other => panic!("unexpected error variant: {:?}", other),
+        }
+    }
+
+    #[test]
+    fn message_converts_into_producer_message() {
+        let mut props = HashMap::new();
+        props.insert("a".to_string(), "1".to_string());
+        props.insert("b".to_string(), "2".to_string());
+
+        let m = Message {
+            payload: b"hello".to_vec(),
+            properties: props.clone(),
+            partition_key: Some("key".into()),
+            ordering_key: Some(vec![1, 2, 3]),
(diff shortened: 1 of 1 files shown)


ChrisWrenDev:
@BewareMyPower Tagging for review - tests from https://github.com/streamnative/pulsar-rs/pull/348

src/producer.rs
```diff
@@ -1205,3 +1205,270 @@ impl<T: SerializeMessage + Sized, Exe: Executor> MessageBuilder<'_, T, Exe> {
...
+                assert_eq!(bm.metadata.partition_key.as_deref(), Some("p"));
+                assert_eq!(bm.metadata.ordering_key.as_deref(), Some(&[7, 8][..]));
+
+                // event time & payload size
+                assert_eq!(bm.metadata.event_time, Some(99));
+                assert_eq!(bm.metadata.payload_size as usize, bm.payload.len());
+            }
+
+            // After draining, no longer full
+            assert!(!batch.is_full().await);
```
BewareMyPower: It's better to put this line immediately after `batch.get_messages().await` because it's a side effect of `get_messages`. (actually the method name is a bit confusing, it should be "drain" rather than "get")

src/producer.rs
```diff
@@ -1205,3 +1205,270 @@ impl<T: SerializeMessage + Sized, Exe: Executor> MessageBuilder<'_, T, Exe> {
...
+                batch.push_back((tx, msg)).await;
+            }
+
+            assert!(
+                batch.is_full().await,
+                "batch should be full at length threshold"
+            );
+
+            // Drain and validate contents -> Batch creates BatchedMessage with mapped fields
+            let drained = batch.get_messages().await;
```
BewareMyPower: We can enhance the assertion by verifying the storage is now empty:

```rust
            assert!(batch.storage.lock().await.storage.is_empty());
```

WDYT?

src/producer.rs
```diff
@@ -1205,3 +1205,270 @@ impl<T: SerializeMessage + Sized, Exe: Executor> MessageBuilder<'_, T, Exe> {
...
+                "batch should be full when total payload meets size threshold"
+            );
+
+            // Drain resets size accounting
+            let drained = batch.get_messages().await;
+            assert_eq!(drained.len(), 2);
+            assert!(
+                !batch.is_full().await,
+                "after drain, batch shouldn't be full"
+            );
```
BewareMyPower: The logic is common that:
1. Verify the batch is full
2. Drain the messages from the batch
3. Verify the batch is now not full

Maybe we can abstract a common method like:

```rust
    async fn drain_messages_from_full_batch(batch: &mut Batch, msg: &str) -> Vec<(oneshot::Sender<Result<CommandSendReceipt, Error>>, BatchedMessage)> {
        assert!(batch.is_full().await, "batch should be full");
        let drained = batch.get_messages().await;
        assert!(
            !batch.is_full().await,
            "after draining, batch shouldn't be full"
        );
        assert!(
            drained.len() > 0,
            "after draining, batch should have messages"
        );
        drained
    }
```

BewareMyPower:
Please also fix the lint errors, you can run these commands locally before pushing new commits:

https://github.com/streamnative/pulsar-rs/blob/520643d12f03ac3a1b9441f71018ca1d59b68155/.github/workflows/rust.yml#L19-L20

ChrisWrenDev:
@BewareMyPower Thanks for the reviews! 

I’ve addressed all the feedback by extracting a shared async helper:
- Added `drain_messages_from_full_batch(&Batch, &str)` used across the batch tests.
- Moved the !is_full() check to immediately after get_messages() (since draining is the side effect).
- Updated the three tests that duplicated this pattern.

I also fixed the assign after default clippy warning.

BewareMyPower (approved):

PR #351, opened 2025-08-27
refactor(producer): remove unnecessary lock contention

ChrisWrenDev:
This PR refactors the batching path in the Pulsar producer to significantly reduce lock contention when multiple tasks are publishing concurrently.

Under load, producers that enable batching contend on the outer Mutex<Batch>. Previously we:
- Held the outer batch lock while serializing all drained messages into a single payload buffer.
- Occasionally performed more work while the lock was held (building receipts, allocating buffers).
This makes other senders wait unnecessarily throttling throughput.

**Changes:**

`TopicProducer::send_batch`:
- Now acquires the outer batch lock only to drain the queued messages (get_messages()), then releases the lock.
- Serialization of batched messages and receipt vector construction happens after the lock is dropped.

`TopicProducer::send_raw`:
- Pushes the new message while holding the lock; if the batch becomes full, drains it under the same lock; then drops the lock.
- Serialization and network send are done outside the critical section.
- If the batch is not full after the push, we return immediately with the message’s SendFuture (no empty batch is sent).

**Flow Changes (before vs. after)**

Before:
1. Acquire outer Mutex<Batch>.
2. Drain messages.
3. Serialize all messages into a buffer while still holding the lock.
4. Build receipts vector.
5. Release lock.
6. Compress & send, then resolve receipts.

After:
1. Acquire outer Mutex<Batch>.
2. Push current message; conditionally drain if full.
3. Release lock.
4. Serialize drained messages into buffer.
5. Compress & send, then resolve receipts.

This change reduces the time the batch lock is held, allowing concurrent senders to continue enqueuing.

**Unchanged Behavior:**
- No public API changes.
- Message ordering within a batch is preserved (we drain in FIFO order).
- Batch semantics (size/byte thresholds) are unchanged.
- We ensure we never send an empty batch (early-return when the batch wasn’t drained).
- Error handling and receipt propagation remain identical; receipts are resolved after the network send completes.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -588,22 +588,22 @@ impl<Exe: Executor> TopicProducer<Exe> {
         match self.batch.as_ref() {
             None => Err(ProducerError::Custom("not a batching producer".to_string()).into()),
             Some(batch) => {
-                let mut payload: Vec<u8> = Vec::new();
-                let mut receipts = Vec::new();
-                let message_count;
-
-                {
-                    let batch = batch.lock().await;
-                    let messages = batch.get_messages().await;
-                    message_count = messages.len();
-                    for (tx, message) in messages {
-                        receipts.push(tx);
-                        message.serialize(&mut payload);
-                    }
-                }
+                let messages = {
+                    let guard = batch.lock().await;
+                    guard.get_messages().await
+                };
+                let message_count = messages.len();
 
                 if message_count == 0 {
                     return Ok(());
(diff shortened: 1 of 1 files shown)


ChrisWrenDev:
@BewareMyPower Tagging for review - lock contention refactor from https://github.com/streamnative/pulsar-rs/pull/348

BewareMyPower (approved):

PR #352, opened 2025-08-29
test: add recv_within helper and timeouts to prevent hangs in consumer tests

ChrisWrenDev:
This PR stabilizes the Pulsar consumer integration tests by ensuring they fail fast instead of hanging indefinitely.

**Changes**
- Introduced a `recv_within` helper that wraps `consumer.next()` with a timeout.
- Updated all consumer tests (`dead_letter_queue`, `dead_letter_queue_batched`, `seek_single_consumer`, etc.) to use the helper instead of waiting unbounded.
- Preserved original test semantics: tests still assert the same conditions, but now exit with a clear timeout error instead of hanging.

**Motivation**
Some tests could previously hang indefinitely if:
- A message was never delivered (e.g. after ack in `consumer_dropped_with_lingering_acks`).
- A seek or DLQ redelivery took longer than expected.
- Pulsar broker startup or CI slowness delayed message flow.
- This caused CI runs to stall without feedback.

**Benefits**
- Prevents infinite hangs in CI and local runs.
- Produces clear failure messages when messages aren’t delivered in time.
- Keeps test runtime bounded and predictable.

### DIFF
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -433,7 +433,7 @@ mod tests {
 
     use futures::{
         future::{select, Either},
-        StreamExt, TryStreamExt,
+        StreamExt,
     };
     use log::LevelFilter;
     use regex::Regex;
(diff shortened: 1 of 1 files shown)


ChrisWrenDev:
@BewareMyPower - Added these fixes after seeing tests hang in other PR https://github.com/streamnative/pulsar-rs/pull/348

BewareMyPower (approved):

PR #353, opened 2025-08-29
ci: harden and optimize rust.yml

ChrisWrenDev:
This PR improves the Rust GitHub Actions workflow for pulsar-rs by making CI runs more reliable, reproducible, and easier to debug.

**Changes**
Stabilized Pulsar startup:
- Explicitly advertise `127.0.0.1` to avoid IPv6/hostname issues
- Added readiness checks (TCP port 6650 + HTTP `/admin/v2/brokers/health`)
- Dump Pulsar logs on failure for better debugging

Improved Cargo usage:
- Added `--workspace --locked` to ensure reproducible builds/tests
- Enforced serial test execution (`RUST_TEST_THREADS=1`) to reduce flakiness

Linting improvements:
- Split Clippy checks into separate feature sets for clarity
- Ensured builds/lints/tests all use the workspace consistently

Toolchain management:
- Use `dtolnay/rust-toolchain` for stable installs of clippy/rustfmt
- Nightly only used if formatting requires it

Workflow hygiene:
- Added `concurrency` to cancel outdated runs
- Set `timeout-minutes` on jobs to prevent hangs
- Updated actions to latest versions (`checkout@v4`, etc.)

**Benefits**
- Fewer flaky test runs across Pulsar versions
- More reproducible builds with locked dependencies
- Faster CI by avoiding redundant toolchain installs
- Clearer lint/test failures with logs and feature set separation


.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
   pull_request:
     branches: [ master ]
 
+concurrency:
+  group: rust-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
ChrisWrenDev, line 11: Adding `concurrency` prevents overlapping runs on the same ref. `cancel-in-progress: true` also auto-cancels superseded pushes to save CI minutes and reduce queue noise.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy,rustfmt
```
ChrisWrenDev, line 21: This ensures the toolchain is explicitly installed and consistent across runners, and that clippy/rustfmt match the compiler version—no ad-hoc rustup calls.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+
+      - name: Clippy (feature set A)
+        run: cargo clippy --tests --features telemetry,protobuf-src -- -D warnings
+
+      - name: Clippy (feature set B)
+        run: cargo clippy --tests --no-default-features --features compression,tokio-rustls-runtime,async-std-rustls-runtime,auth-oauth2,telemetry,protobuf-src -- -D warnings
+
       - name: Check format
-        run: cargo +nightly fmt --all --check
+        run: cargo fmt --all --check
```
ChrisWrenDev, line 34: Switched formatting to the stable toolchain for determinism and less churn. It also should avoid nightly component breakage.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
     branches: [ master ]
 
+concurrency:
+  group: rust-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
```
ChrisWrenDev, line 16: Hard caps for hanging or stalled runs to free runners and keep CI predictable.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+
       - name: Build
-        run: cargo build --features protobuf-src
-      - name: Check Clippy
-        run: |
-          cargo clippy --tests --features telemetry,protobuf-src -- -D warnings
-          cargo clippy --tests --no-default-features --features compression,tokio-rustls-runtime,async-std-rustls-runtime,auth-oauth2,telemetry,protobuf-src -- -D warnings
-      - name: Install nightly rustfmt
-        run: rustup toolchain install nightly --component rustfmt
+        run: cargo build --features protobuf-src --locked
```
ChrisWrenDev: Adding `--locked` enforces using Cargo.lock for reproducible dependency resolution and fails fast if the lockfile is out of date.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+
+      - name: Clippy (feature set A)
+        run: cargo clippy --tests --features telemetry,protobuf-src -- -D warnings
+
+      - name: Clippy (feature set B)
+        run: cargo clippy --tests --no-default-features --features compression,tokio-rustls-runtime,async-std-rustls-runtime,auth-oauth2,telemetry,protobuf-src -- -D warnings
```
ChrisWrenDev, line 31: Split clippy into two steps makes failures easier to pinpoint and keeps logs scoped to the feature combination that failed.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+
       - name: Check format
-        run: cargo +nightly fmt --all --check
+        run: cargo fmt --all --check
 
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     strategy:
+      fail-fast: false
```
ChrisWrenDev, line 40: Adding this should run all matrix entries even if one fails, so we see regressions across all tested Pulsar versions in a single run.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+      - name: Start Pulsar Standalone Container
+        run: |
+          docker run --name pulsar \
+            -p 6650:6650 -p 8080:8080 \
+            -d \
+            -e GITHUB_ACTIONS=true -e CI=true \
+            -e PULSAR_PREFIX_advertisedAddress=127.0.0.1 \
+            -e PULSAR_PREFIX_advertisedListeners=pulsar://127.0.0.1:6650 \
+            -e PULSAR_PREFIX_brokerServicePort=6650 \
+            -e PULSAR_PREFIX_webServicePort=8080 \
```
ChrisWrenDev, line 57: These PULSAR_PREFIX force the broker to advertise 127.0.0.1 and explicit ports, which will help avoid connection issues when the container’s hostname/IP would otherwise be advertised.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+          (echo > /dev/tcp/127.0.0.1/6650) >/dev/null 2>&1 || { echo "Port 6650 not ready"; exit 1; }
+
+          echo "Waiting for HTTP health..."
+          for i in {1..90}; do
+            if curl -sf http://127.0.0.1:8080/admin/v2/brokers/health >/dev/null; then
+              echo "Broker health OK"
+              exit 0
+            fi
+            sleep 1
+          done
```
ChrisWrenDev, line 81: Wait for Pulsar readiness eliminates race conditions by waiting for port 6650 and `/admin/v2/brokers/health` before running tests. It also prints recent logs on timeout for faster debugging.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+
+          echo "Health endpoint didn't respond in time:"
+          docker logs --tail=200 pulsar || true
+          exit 1
+
+      - name: Run tests (serial, verbose)
+        env:
+          RUST_BACKTRACE: 1
+          RUST_LOG: pulsar=debug
+          RUST_TEST_THREADS: 1
```
ChrisWrenDev, line 91: These serialize integration tests to reduce flakiness against a single standalone broker and increases diagnostics on failures.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+        run: cargo test --features protobuf-src --locked -- --nocapture
+
+      - name: Dump Pulsar logs on failure
+        if: failure()
+        run: |
+          echo "==== PULSAR CONTAINER LOGS ===="
+          docker logs pulsar || true
```
ChrisWrenDev, line 98: Dump Pulsar logs on failure automatically includes broker logs in the job output to speed up root-cause analysis when tests fail.

.github/workflows/rust.yml
```diff
@@ -6,32 +6,98 @@ on:
...
+
+      - name: Stop Pulsar
+        if: always()
+        run: docker rm -f pulsar || true
```
ChrisWrenDev, line 102: Always stopping the Pulsar container at the end ensures clean teardown so subsequent jobs can reuse the container name and we don’t leak resources on the runner.

Cargo.lock
```diff
@@ -0,0 +1,3693 @@
+# This file is automatically @generated by Cargo.
```
BewareMyPower: From the discussion here: https://github.com/rust-lang/cargo/issues/315

>  it should be ignored for libraries, tracked for binaries.

We should keep ignoring it for libraries. You can also see tokio ignores it as well: https://github.com/tokio-rs/tokio/blob/master/.gitignore#L2

BewareMyPower (changes requested):

Cargo.lock
ChrisWrenDev: I've removed Cargo.lock and the --locked flag

BewareMyPower (approved):

PR #354, opened 2025-09-03
feat: refactor producer batching with timeout supported

### LINKED ISSUE #306: Missing `batchingMaxPublishDelay` option
In Java client, there is a `batchingMaxPublishDelay` option, which is 1 millisecond by default:

https://github.com/apache/pulsar/blob/02147454c425b92f0cd1caefa73b9339db6a0269/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ProducerBuilderImpl.java#L245

```java
    public ProducerBuilder<T> batchingMaxPublishDelay(long batchDelay, @NonNull TimeUnit timeUnit) {
        conf.setBatchingMaxPublishDelayMicros(batchDelay, timeUnit);
        return this;
    }
```

And:

```java
    @ApiModelProperty(
            name = "batchingMaxPublishDelayMicros",
            value = "Batching time period of sending messages."
    )
    private long batchingMaxPublishDelayMicros = TimeUnit.MILLISECONDS.toMicros(1);
```

Without such an option, if we set the `batch_size` to 1000 in pulsar-rs, and there are only 500 messages to be sent, then no messages will be sent to the Pulsar broker side at all:

```rust
use pulsar::{Authentication, ProducerOptions, Pulsar, TokioExecutor};

#[tokio::main]
    env_logger::init();

    let addr = "pulsar://localhost:6650".to_string();
    let mut builder = Pulsar::builder(addr, TokioExecutor);

    let pulsar: Pulsar<_> = builder.build().await?;
    let mut producer = pulsar
        .producer()
        .with_topic("persistent://public/default/test-batch")
        .with_options(ProducerOptions {
            encrypted: None,
            metadata: Default::default(),
            schema: None,
            batch_size: Some(1000),
            batch_byte_size: Some(128 * 1024),
            compression: None,
            access_mode: None,
        })
        .build()
        .await?;

    let mut counter = 0usize;
    loop {
        producer
            .send(format!("Hello-{}", counter))
            .await
            .unwrap();

        counter += 1;
        println!("{counter} messages");

        if counter >= 500 {
            producer.close().await.expect("Unable to close connection");
            break;
        }
    }

    Ok(())
}
```

It looks like there is a buffer in the backend, and only when the buffer is full(size == batch_size), the pulsar-rs will do the actual sending work.?

I think the batch sending needs to be triggered when one of the below conditions is met:

1. the buffer is full

2. a given timeout, such as 1 millisecond

BewareMyPower:
Fixes https://github.com/streamnative/pulsar-rs/issues/306

### Motivation

Without batching timeout, the batching feature will be hard to use because the messages won't be sent if the size of byte size limit is not reached. Users can call `send_batch` to flush all messages, but awaiting this method will wait until the batched send is done. Without awaiting `send_batch`, the producer will be borrowed and cannot send next message.

In addition, the current batching implementation has some minor issues:
- The returned message id by `send_non_blocking` does not have batch index set
- If `batch_size` is not set, even if `max_byte_size` is set, the message won't be batched.

### Modifications

Add a producer new option `batch_timeout` to specify whether to enable the batch timeout. Refactor the implementation with the message delivery based design. When batching is enabled, create two tasks:
1. `batch_process_loop`
   - Receive all single messages (*) from the producer and add them to batch and flush the batch (all messages in the `BatchStorage`) as a whole to `message_send_loop`
   - If batch timeout is configured, when the 1st message is added, start a delayed task with the given timeout. If the timeout is reached first, flush the messages regardless of the size limit.
2. `message_send_loop`: receive batches and send batches to broker one by one

A single messages is composed of an `oneshot::Sender` and a message (`BatchedMessage`). The future returned by `send_non_blocking(...).await` is the `oneshot::Receiver` associated to that sender. When `send_non_blocking` is called, create an `oneshot::channel` and return the receiver as the send future. The sender and the user provided message will be wrapped to a `SingleMessage` and then sent to `batch_process_loop`.

Specifically, when `send_batch` is called, send only a `oneshot::Sender` to the `batch_process_loop` to flush existing messages and wait the result in `send_batch`.

Handle closing producer carefully by closing the sender to trigger the exit of `batch_process_loop`. When it exits, fail all pending sends with a newly added error code `ProducerError::Closed`.

### Tests

Improve `batching` to cover various cases:
1. Only `batch_size` is set.
2. Only `batch_byte_size` is set.
3. Only `batch_timeout` is set.
4. All these configs are set.

Add a new `flush` test to cover the semantic of `send_batch` and the behavior after closing the producer.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -11,11 +11,10 @@ use std::{
 };
 
 use futures::{
-    channel::oneshot,
-    future::try_join_all,
-    lock::Mutex,
+    channel::{mpsc, oneshot},
+    future::{self, try_join_all, Either},
     task::{Context, Poll},
-    Future,
+    Future, SinkExt, StreamExt,
 };
 
 use crate::{
@@ -149,11 +148,23 @@ pub struct ProducerOptions {
     /// batch size in bytes treshold (only relevant when batch_size active).
     /// batch is sent when batch size in bytes is reached
     pub batch_byte_size: Option<usize>,
+    /// the batch will be sent if this timeout is reached after the 1st message is added into the
+    /// batch even if it does not reach the size or byte size limit.
+    pub batch_timeout: Option<Duration>,
     /// algorithm used to compress the messages
     pub compression: Option<Compression>,
     /// producer access mode: shared = 0, exclusive = 1, waitforexclusive =2,
     /// exclusivewithoutfencing =3
     pub access_mode: Option<i32>,
+}
+
+impl ProducerOptions {
+    fn enabled_batching(&self) -> bool {
+        match self.batch_size {
+            Some(batch_size) => batch_size > 1,
+            None => self.batch_byte_size.is_some() || self.batch_timeout.is_some(),
+        }
+    }
 }
 
 /// Wrapper structure that manges multiple producers at once, creating them as needed
@@ -502,12 +513,12 @@ struct TopicProducer<Exe: Executor> {
     id: ProducerId,
     name: ProducerName,
     topic: String,
-    message_id: SerialId,
-    // putting it in a mutex because we must send multiple messages at once
-    // while we might be pushing more messages from elsewhere
-    batch: Option<Mutex<Batch>>,
+    sequence_id: SerialId,
     compression: Option<Compression>,
     options: ProducerOptions,
+    // When batching is enabled, `send_non_blocking` will send the batch to the
+    // `batch_process_loop` task instead of sending it directly.
+    batch_msg_sender: Option<mpsc::Sender<Option<SingleMessage>>>,
 }
 
 impl<Exe: Executor> TopicProducer<Exe> {
@@ -523,11 +534,9 @@ impl<Exe: Executor> TopicProducer<Exe> {
 
         let topic = topic.into();
         let producer_id = PRODUCER_ID_GENERATOR.fetch_add(1, Ordering::SeqCst);
-        let sequence_ids = SerialId::new();
+        let sequence_id = SerialId::new();
 
         let topic = topic.clone();
-        let batch_size = options.batch_size;
-        let batch_byte_size = options.batch_byte_size;
         let compression = options.compression.clone();
         let mut connection = client.manager.get_connection(&addr).await?;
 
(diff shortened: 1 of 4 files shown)


BewareMyPower:
@claude Please help review this PR

src/producer.rs
```diff
@@ -965,111 +949,179 @@ impl<Exe: Executor> ProducerBuilder<Exe> {
...
+                    recv_future = msg_receiver.next();
+                }
             },
-            payload: message.payload,
-        };
-        self.storage.lock().await.push_back(tx, batched)
+            Either::Left((None, _)) => {
+                debug!("batch_process_loop: channel closed, exiting");
+                break;
+            }
```
ChrisWrenDev: Should we be flushing any pending messages here to prevent leaving all oneshot senders unresolved?

Example:
```rust
if !batch_storage.storage.is_empty() {
  let _ = batch_sender.clone().send(batch_storage.get_messages()).await;
}
```

src/producer.rs
```diff
@@ -585,257 +627,199 @@ impl<Exe: Executor> TopicProducer<Exe> {
...
-
-                trace!("sending a batched message of size {}", message_count);
-                let send_receipt = self.send_compress(message).await?.await.map_err(Arc::new);
-                for resolver in receipts {
-                    let _ = resolver.send(
-                        send_receipt
-                            .clone()
-                            .map_err(|e| ProducerError::Batch(e).into()),
-                    );
-                }
```
ChrisWrenDev, line 623: This change to `send_batch` breaks users who rely on `send_batch().await` as a flush barrier (tests, graceful shutdown). It could be restored to blocking behavior (return a future that resolves once the batch is acked) or clearly rename this as `trigger_flush()` and add a new blocking `flush().await`.

src/producer.rs
```diff
@@ -965,111 +949,179 @@ impl<Exe: Executor> ProducerBuilder<Exe> {
...
+                }
+
+                let fail = |receipts: Vec<oneshot::Sender<Result<CommandSendReceipt, Error>>>,
+                            description: &str,
+                            e: Error| {
+                    let error = Arc::new(Error::Custom(format!("{description}: {e}")));
+                    for tx in receipts {
+                        let _ = tx.send(Err(Error::Producer(ProducerError::Batch(error.clone()))));
+                    }
+                };
```
ChrisWrenDev: This will discard original variants (e.g. ConnectionError), are we able to preserve structured errors so callers can distinguish between errors (e.g transport vs compression failures).

src/producer.rs
```diff
@@ -965,111 +949,179 @@ impl<Exe: Executor> ProducerBuilder<Exe> {
...
+                    } else {
+                        timer_future = match batch_storage.timeout {
+                            Some(timeout) if batch_storage.storage.len() == 1 => {
+                                Box::pin(executor.delay(timeout))
+                            }
+                            _ => previous_timer_future,
+                        };
+                    }
+                    recv_future = msg_receiver.next();
+                }
```
ChrisWrenDev: It looks like there could be a potential race if `batch_timeout` fires at the same time as a `send_batch()` call which would cause both branches to try flush the same batch.

If we only flush if storage is non-empty and always cancel the timer after a flush it should prevent it.

src/producer.rs
BewareMyPower: It makes sense, let me try fixing it.

src/producer.rs
BewareMyPower: It does not make sense because when `batch_process_loop` exits, it means the producer is dropped or `close` is called (with my latest implementation). In this case, `message_send_loop` will exit as well, so flushing them won't help.

I'd like to add a warning here if there were some messages not sent. Please note that in my latest implementation, `close` will wait until the `batch_process_loop` exits.

src/producer.rs
BewareMyPower: ~~Ah yes, I'd check all places that ignore the result of `tx.send` (actually I added some checks before)~~

Yes, I made this change because `Error` is not cloneable. But here we can use `Arc::new(e)` instead.

src/producer.rs
BewareMyPower: Good suggestion, it makes sense

src/producer.rs
BewareMyPower: BTW, I have to introduce `RefCell` since `close` accepts an immutable reference.

BewareMyPower: Now I failed all pending send futures after close

BewareMyPower:
@ChrisWrenDev Thanks for your valuable reviews. All comments should be addressed now, PTAL again

freeznet (approved):

PR #359, opened 2025-09-21
fix: restore the Send trait for Producer

### LINKED ISSUE #357: Regression: Producer is no longer Send in 6.4.0 (RefCell in Batch breaks Sync/Send`)
After upgrading **pulsar-rs** from `6.3.1` to `6.4.0`, code that previously worked with `tokio::spawn` now fails to compile due to `Producer` no longer being `Send`.

Error message:

```
error: `RefCell<std::option::Option<producer::Batch>>` cannot be shared between threads safely
   |
128 | /         tokio::spawn(async move {
129 | |             let producer = pulsar
130 | |                 .producer()
131 | |                 .with_topic(topic)
...   |
147 | |             Self::start(cfg_entry, producer, rx).await;
148 | |         });
    | |__________^
    |
    = help: within `pulsar::Producer<TokioExecutor>`, the trait `Sync` is not implemented for `RefCell<std::option::Option<producer::Batch>>`
    = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock` instead
```

It seems that in **6.4.0**, the internal implementation of `Producer` introduced a `RefCell<Option<Batch>>` inside `TopicProducer`, which makes `Producer<TokioExecutor>` not `Send` anymore.

This is a breaking change compared to `6.3.1`, where `Producer` was `Send` and could be moved into `tokio::spawn`.

BewareMyPower:
Fixes #357

Change `TopicProducer::close` to accept a mutable reference and remove the `pub` modifier. It's reasonable because `TopicProducer` should be an inner struct that should never be accessed from applications. The immutable reference does not make sense because `Producer::close` also accepts a mutable reference and this method actually changes the internal state (e.g. via a `CloseProducer` RPC).

After that, remove the use of `RefCell` and add `move_producer_to_spawned_task` test to avoid regression.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -1507,7 +1507,6 @@ mod tests {
         }
     }
 
-
     #[tokio::test]
     async fn move_producer_to_spawned_task() {
         let _result = log::set_logger(&TEST_LOGGER);
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -735,7 +735,7 @@ impl<Exe: Executor> TopicProducer<Exe> {
     }
 
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
-    pub async fn close(&mut self) -> Result<(), Error> {
+    async fn close(&mut self) -> Result<(), Error> {
         match &mut self.batch {
             None => {
                 self.connection.sender().close_producer(self.id).await?;
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -1,6 +1,5 @@
 //! Message publication
 use std::{
-    cell::RefCell,
     collections::{BTreeMap, HashMap, VecDeque},
     io::Write,
     pin::Pin,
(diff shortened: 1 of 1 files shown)


BewareMyPower:
The root cause to introduce `RefCell` is that `TopicProducer::close` accepts a immutable reference https://github.com/streamnative/pulsar-rs/blob/51b93edc0cda36eb76319b77ec37be01d063d09e/src/producer.rs#L745

and it's a public method.

However, `TopicProducer` is a struct in a non-public enum

https://github.com/streamnative/pulsar-rs/blob/51b93edc0cda36eb76319b77ec37be01d063d09e/src/producer.rs#L514-L517

while the `Producer::close` method, which is called by the user directly, accepts a mutable reference

https://github.com/streamnative/pulsar-rs/blob/51b93edc0cda36eb76319b77ec37be01d063d09e/src/producer.rs#L504-L506

so I'd assume changing `TopicProducer::close` from `close(&self)` to `close(&mut self)` is valid and won't cause any breaking change. Based on the assumption, the code would be much more natural to write.

FlorentinDUBOIS (approved):

Issue #344, opened 2025-07-29
Provide a feature to use `ring` instead of `aws-lc-sys`


Logarithmus:
`aws-lc-sys` is slow to compile, and most people don't care about FIPS or whatever, so `ring` is better in most cases.
You need to put `default-features = false` for `rustls`, as simple as that.

f1shl3gs:
This annoying me too, because rustls does not allow users enable `ring` and `aws_lc_rs` at the same time. So adding pulsar will panic programe once TLS is used.

See  
https://github.com/rustls/rustls/blob/e2f6115548b54e15ec52517208e4b4c6f27aa7db/rustls/src/crypto/mod.rs#L247
https://github.com/rustls/rustls/blob/e2f6115548b54e15ec52517208e4b4c6f27aa7db/rustls/src/crypto/mod.rs#L279

PR #362, opened 2025-09-21
Allow to change rustls crypto provider

dsferruzza:
Hi!

I had the issue explained in #344 (and it was blocking me because `aws-lc-sys` does not seem to be able to easily cross compile to ARMv6) so I made this change and it seems to fix it.
I'm using my fork successfully in my app.

I'm not sure if this is the right way to fix this though: if I understand correctly, this relies on having either the `aws-lc-rs` or the `ring` Rustls features enabled in the dependency tree(?)

Another way could be to introduce the following features in this crate:

- `tokio-rustls-runtime-ring`
- `tokio-rustls-runtime-aws-lc`
- `async-std-rustls-runtime-ring`
- `async-std-rustls-runtime-aws-lc`

Each would enable the correct features in the Rustls-related dependencies.
The `tokio-rustls-runtime` and `async-std-rustls-runtime` features would be aliases to whatever should be the default.

What do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ crc = "^3.3.0"
 data-url = { version = "^0.3.1", optional = true }
 flate2 = { version = "^1.1.1", optional = true }
 futures = "^0.3.31"
-futures-rustls = { version = "^0.26.0", optional = true } # replacement of crate async-rustls (also a fork of tokio-rustls)
+futures-rustls = { version = "^0.26.0", default-features = false, features = ["tls12", "logging"], optional = true } # replacement of crate async-rustls (also a fork of tokio-rustls)
 log = "^0.4.27"
 lz4 = { version = "^1.28.0", optional = true }
 native-tls = { version = "^0.2.12", optional = true }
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -39,13 +39,13 @@ prost = "^0.13.4"
...
 regex = "^1.11.1"
-rustls = { version = "^0.23.27", optional = true }
+rustls = { version = "^0.23.27", default-features = false, optional = true }
 snap = { version = "^1.1.1", optional = true }
 serde = { version = "^1.0.219", features = ["derive"], optional = true }
 serde_json = { version = "^1.0.140", optional = true }
 tokio = { version = "^1.45.0", features = ["rt", "net", "time"], optional = true }
 tokio-util = { version = "^0.7.15", features = ["codec"], optional = true }
-tokio-rustls = { version = "0.26.2", optional = true }
+tokio-rustls = { version = "0.26.2", default-features = false, optional = true }
```
BewareMyPower: Since it disables the default features, will it bring any breaking change for TLS use cases? Currently there is no integration test for TLS, so it's not quick to verify it.

src/consumer/engine.rs
```diff
@@ -433,7 +433,7 @@ impl<Exe: Executor> ConsumerEngine<Exe> {
     async fn process_payload(
         &mut self,
         message: CommandMessage,
-        mut payload: Payload,
+        #[allow(unused_mut)] mut payload: Payload,
```
BewareMyPower: It seems to be an irrelevant change (as well as the changes in `producer.rs`)?

dsferruzza: This was indeed not relevant, I removed it.

Cargo.toml
dsferruzza: I took some time to reimplement this in a way that should not break compatibility. Does it look better?

BewareMyPower (approved):

Issue #322, opened 2024-08-14
Support single/multi partition routing modes


rkrishn7:
Relates to #150 

Currently, the default and only supported routing mode for partitioned producers appears to be round robin. We should support single/multi partition routing modes as well.

More info: https://pulsar.apache.org/docs/3.3.x/concepts-messaging/#routing-modes

PR #366, opened 2025-10-06
feat: Updating producer to configure RoutingPolicy - adding partition_key as the default for Round Robin if set

kaustubh-garimella-genius:
Should address: https://github.com/streamnative/pulsar-rs/issues/322

https://pulsar.apache.org/docs/next/client-libraries-producers/#choose-partitions-when-using-a-key

Copying the behavior in the Java Client, where the producer takes in a configuration for a RoutingPolicy (RoundRobin, Single or Custom) for publishing to a partitioned topic. 

Non breaking behavior since if you don't specify a routing policy, it will automtically be Round Robin. Similarly to the Java Client, if you explicitly have a partition_key set in your message, it will deterministically hash the key and route to a partition that way

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,18 +18,28 @@ keywords = ["pulsar", "api", "client"]
 [dependencies]
 async-channel = "^2.3.1"
 async-trait = "^0.1.88"
-async-std = { version = "^1.13.1", features = ["attributes", "unstable"], optional = true }
+async-std = { version = "^1.13.1", features = [
+    "attributes",
+    "unstable",
+], optional = true }
 async-native-tls = { version = "^0.5.0", optional = true }
 asynchronous-codec = { version = "^0.7.0", optional = true }
 bytes = "^1.9.0"
-chrono = { version = "^0.4.41", default-features = false, features = ["clock", "std"] }
+chrono = { version = "^0.4.41", default-features = false, features = [
+    "clock",
+    "std",
+] }
 crc = "^3.3.0"
 data-url = { version = "^0.3.1", optional = true }
 flate2 = { version = "^1.1.1", optional = true }
 futures = "^0.3.31"
-futures-rustls = { version = "^0.26.0", default-features = false, features = ["tls12", "logging"], optional = true } # replacement of crate async-rustls (also a fork of tokio-rustls)
+futures-rustls = { version = "^0.26.0", default-features = false, features = [
+    "tls12",
+    "logging",
+], optional = true } # replacement of crate async-rustls (also a fork of tokio-rustls)
 log = "^0.4.27"
 lz4 = { version = "^1.28.0", optional = true }
+murmur3 = "0.5"
 native-tls = { version = "^0.2.12", optional = true }
 nom = { version = "^7.1.3", default-features = false, features = ["alloc"] }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -18,18 +18,28 @@ keywords = ["pulsar", "api", "client"]
...
+] }
 crc = "^3.3.0"
 data-url = { version = "^0.3.1", optional = true }
 flate2 = { version = "^1.1.1", optional = true }
 futures = "^0.3.31"
-futures-rustls = { version = "^0.26.0", default-features = false, features = ["tls12", "logging"], optional = true } # replacement of crate async-rustls (also a fork of tokio-rustls)
+futures-rustls = { version = "^0.26.0", default-features = false, features = [
+    "tls12",
+    "logging",
+], optional = true } # replacement of crate async-rustls (also a fork of tokio-rustls)
```
BewareMyPower: Please revert unnecessary format changes on `Cargo.toml`, which would be unclear for the actual dependency changes.

kaustubh-garimella-genius: Updated 👍

kaustubh-garimella-genius:
@BewareMyPower when you get a chance,  please take another look. Updated with some additional tests and slightly changed the behavior to work with a stable vec of producers.

kaustubh-garimella-genius:
Fixed the round robin unit test

src/routing_policy.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use murmur3::murmur3_32;
+
+use crate::producer::Message;
+
+#[derive(Clone, Default)]
+pub enum RoutingPolicy {
+    #[default]
+    RoundRobin,
+    Single(usize),
```
BewareMyPower: This strategy is different from other clients. The index should be generated once rather than provided by the application side. Because the application cannot know the actual number of partitions in advance, the index could exceed the actual number of partitions

BewareMyPower (changes requested):

src/producer.rs
```diff
@@ -1535,4 +1563,173 @@ mod tests {
...
+                pulsar.clone(),
+                addr.clone(),
+                format!("{}-{}", topic.clone(), i),
+                Some(i.to_string()),
+                options.clone(),
+            )
+            .await
+            .unwrap();
+            producers.push(producer);
+        }
```
BewareMyPower: It's better to leverage the REST API to create a partitioned topic. See example here: https://github.com/streamnative/pulsar-rs/blob/d1f18e6ef8dbd1bc30a82bafc9875d842e27dcf6/src/lib.rs#L816

https://pulsar.apache.org/docs/next/admin-api-topics/#create-1

Then you won't need to create the `PartitionedProducer` directly for tests, whose visibility is only within the crate.

src/routing_policy.rs
kaustubh-garimella-genius: 👍 Updated to be generated once

src/producer.rs
kaustubh-garimella-genius: Makes sense, was trying to isolate to just that behavior but refactored to use admin api to generate partitioned topic

BewareMyPower (approved):

BewareMyPower:
Please fix lint error

BewareMyPower (approved):

PR #370, opened 2025-10-27
feat: enhance consumer functionality tests

crossoverJie:
This pull request enhances the consumer integration tests by adding more comprehensive checks for consumer statistics, acknowledgment behaviors, and resource cleanup. It also improves the clarity and reliability of the tests by explicitly naming consumers and verifying their connections.

### DIFF
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -747,6 +747,9 @@ mod tests {
                 "waiting for a message should have timed out, since we already acknowledged the \
                  only message in the queue"
             );
+            // clean up
+            consumer.unsubscribe().await.unwrap();
+            consumer.close().await.unwrap();
         }
     }
 
@@ -1121,10 +1124,6 @@ mod tests {
         // then check if all messages were received
         assert_eq!(50, consumed_1);
         assert_eq!(100, consumed_2);
-
-        // clean up
-        consumer_1.unsubscribe().await.unwrap();
-        consumer_1.close().await.unwrap();
     }
 
     #[tokio::test]
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -557,6 +557,7 @@ mod tests {
             .with_subscription("consumer_1")
             .with_consumer_name("consumer_1")
             .with_topics([&topic1, &topic2])
+            .with_subscription_type(SubType::Exclusive)
             .build()
             .await
             .unwrap();
@@ -564,6 +565,7 @@ mod tests {
         assert!(&consumer_1.check_connection().await.is_ok());
 
         let mut consumer_2: Consumer<TestData, _> = builder
+            .clone()
             .with_subscription("consumer_2")
             .with_topic_regex(Regex::new(&format!("multi_consumer_[ab]_{topic_n}")).unwrap())
             .build()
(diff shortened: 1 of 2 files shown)


crossoverJie:
@BewareMyPower PTAL

src/consumer/mod.rs
```diff
@@ -593,6 +605,48 @@ mod tests {
...
+            if i == 1 {
+                latest_msg = Some(msg.unwrap());
+            }
+        }
+
+        consumer_1
+            .cumulative_ack(&latest_msg.unwrap())
+            .await
+            .unwrap();
+        let msg = recv_within(&mut consumer_1, DEFAULT_RECV_TIMEOUT).await;
```
BewareMyPower: It would be better to reduce the timeout here to reduce test time.

A better test pattern to verify ack works is to send a new message and recreate the consumer to consume it.

src/consumer/mod.rs
```diff
@@ -593,6 +605,48 @@ mod tests {
...
+        assert!(msg.is_err(), "timed out waiting for next() after 5s");
+
+        // cleanup
+        for consumer in [&mut consumer_1, &mut consumer_2].iter_mut() {
+            consumer.unsubscribe().await.unwrap();
```
BewareMyPower, line 664: To verify `unsubscribe` works as expected, you should create another consumer with the same subscription name and see if it would fail (the subscription type should be changed to `Exclusive`)

src/consumer/mod.rs
```diff
@@ -593,6 +605,48 @@ mod tests {
...
+        ])
+        .await
+        .unwrap();
+        let mut latest_msg = None;
+        for i in 0..2 {
+            let msg = recv_within(&mut consumer_1, DEFAULT_RECV_TIMEOUT).await;
+            println!("consumer_1 receive {}: {:?}", i, msg);
+            if i == 1 {
+                latest_msg = Some(msg.unwrap());
+            }
```
BewareMyPower: The `i == 1` check seems unnecessary?

src/consumer/mod.rs
crossoverJie: When a consumer is created and then a new message is sent, the consumer will by default use the [latest message](https://github.com/streamnative/pulsar-rs/blob/f2cbcf1e0fe55508476f4ca48f1e26581e66ad75/src/consumer/initial_position.rs#L8) consumption information, which does not verify that previous messages have been acknowledged.

src/consumer/mod.rs
crossoverJie: Done

src/consumer/mod.rs
BewareMyPower: You can change it to `Earliest` to test.

Currently the test logic is:
1. send msg-0, msg-1
2. ack msg-1
3. receive with 5 seconds

The suggested test logic is:
1. send msg-0, msg-1
2. ack msg-1
3. send msg-2
4. close the consumer and recreate it
5. verify msg-2 will be received

Then you won't need to wait for a timeout. Additionally, if the ack failed, the 1st message won't be msg-2 no matter what the initial position is. (earliest -> msg-0, latest -> timeout)

crossoverJie: You are right, it has been modified.

> You can change it to `Earliest` to test.


I didn't notice that it had been set to [Earliest](https://github.com/streamnative/pulsar-rs/blob/a632501a4d299ce4e9dd91e80d73ba9b29b255b9/src/consumer/mod.rs#L551).

BewareMyPower (approved):

PR #371, opened 2025-10-28
feat: add receiver queue size configuration to consumer

### LINKED ISSUE #365: Missing configurable receiver_queue_size option for consumer
Other clients have a receiverQueueSize option, e.g. [C++](https://pulsar.apache.org/api/cpp/3.7.x/classpulsar_1_1_consumer_configuration.html#a265d2cd1e9d1d329eff9b98346f245c2). 

As far as I can tell, there is no analog to that in this client. This seems to be the mpsc channel used in `consumer/topic.rs` on line 106 that is hardcoded to 1000. Are there any plans on implementing this? What would complete support entail? 

This can have significant impact on shared topics with multiple subscribers.

crossoverJie:
fix https://github.com/streamnative/pulsar-rs/issues/365

Add support for configuring the receiver queue size for consumers. It introduces a new `receiver_queue_size` option to the consumer builder and configuration, exposes it via the public API, and includes tests to verify the new functionality.

### DIFF
--- a/src/consumer/config.rs
+++ b/src/consumer/config.rs
@@ -22,8 +22,6 @@ pub struct ConsumerConfig {
     pub(crate) consumer_name: Option<String>,
     /// numerical id of the consumer
     pub(crate) consumer_id: Option<u64>,
-    /// size of the receiver queue
-    pub(crate) receiver_queue_size: Option<u32>,
     /// time after which unacked messages will be sent again
     pub(crate) unacked_message_redelivery_delay: Option<Duration>,
     /// consumer options
--- a/src/consumer/config.rs
+++ b/src/consumer/config.rs
@@ -22,6 +22,8 @@ pub struct ConsumerConfig {
     pub(crate) consumer_name: Option<String>,
     /// numerical id of the consumer
     pub(crate) consumer_id: Option<u64>,
+    /// size of the receiver queue
+    pub(crate) receiver_queue_size: Option<u32>,
     /// time after which unacked messages will be sent again
     pub(crate) unacked_message_redelivery_delay: Option<Duration>,
     /// consumer options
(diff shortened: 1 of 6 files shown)


crossoverJie:
@BewareMyPower PTAL

src/consumer/config.rs
```diff
@@ -22,6 +22,8 @@ pub struct ConsumerConfig {
     pub(crate) consumer_name: Option<String>,
     /// numerical id of the consumer
     pub(crate) consumer_id: Option<u64>,
+    /// size of the receiver queue
+    pub(crate) receiver_queue_size: Option<u32>,
```
BewareMyPower: This field should be added to `ConsumerOptions`, then we don't need to add a new `receiver_queue_size` method

src/consumer/topic.rs
```diff
@@ -103,7 +104,8 @@ impl<T: DeserializeMessage, Exe: Executor> TopicConsumer<T, Exe> {
                 return Err(Error::Executor);
             }
         }
-        let (tx, rx) = mpsc::channel(1000);
+        let receiver_queue_size = receiver_queue_size.unwrap_or(1000);
```
BewareMyPower: What if the receiver queue size is 0? Will the behavior be correct?

crossoverJie: Since `zero_queue_consumer` is not currently implemented, when `receiver_queue_size` is set to 0, it will default to an initial value of 1000.

src/consumer/config.rs
crossoverJie: Done

src/consumer/options.rs
```diff
@@ -76,4 +78,10 @@ impl ConsumerOptions {
         self.initial_position = initial_position;
         self
     }
+
+    #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    pub fn with_receiver_queue_size(mut self, size: u32) -> Self {
+        self.receiver_queue_size = Some(if size == 0 { 1000 } else { size });
```
BewareMyPower, line 85: Could you add a test for `size == 0` case as well as a TODO for handling this case? It would also be better opening an issue

crossoverJie: Done

BewareMyPower (approved):

PR #380, opened 2025-12-30
perf: remove unnecessary clones from multitopic producer

mattsre:
Small improvement to `send_non_blocking()` on multi-topic producers to use the entry API. Removes the need to clone the topic name and get the producer after inserting it. Additional removal of an unnecessary clone on the `connect()` method

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -1,6 +1,6 @@
 //! Message publication
 use std::{
-    collections::{BTreeMap, HashMap, VecDeque},
+    collections::{btree_map::Entry, BTreeMap, HashMap, VecDeque},
     io::Write,
     pin::Pin,
     sync::{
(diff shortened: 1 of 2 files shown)


src/producer.rs
```diff
@@ -245,20 +245,22 @@ impl<Exe: Executor> MultiTopicProducer<Exe> {
...
-                .with_options(self.options.clone());
-            if let Some(name) = &self.name {
-                builder = builder.with_name(name.clone());
+        let producer = match self.producers.entry(topic) {
+            Entry::Vacant(entry) => {
+                let mut builder = self
+                    .client
+                    .producer()
+                    .with_topic(entry.key())
+                    .with_options(self.options.clone());
```
mattsre, line 254: The multi-topic producer should probably share options with the inner producers to remove this clone. Would you be open to me referencing the the producer options from the multi-topic producer instead of cloning for each topic in the multi-producer? I'd need to explore how to do that without changing the single-topic producer too much

BewareMyPower: IMO, the core optimization of this PR is reduce two queries on a tree map (`contains_key` and `get_mut`) to one (`entry`) on the hot path `send_non_blocking`. But the case when the producer is absent should not be frequent, cloning the options is acceptable.

BewareMyPower: And yes, sharing the same options makes sense.

BewareMyPower (approved):

PR #384, opened 2026-01-09
fix: Correctly handle consumer errors and removal due to pulsar cluster restarts by unconditionally update_topics in MultiTopicConsumer

chamons:
While testing https://github.com/streamnative/pulsar-rs/pull/379 it was found in that cases when the entire pulsar broker was restarted, we'd hit a similar case:

- Instead of hitting Some(Err(e) at https://github.com/streamnative/pulsar-rs/blob/a14e8a15144a48d7d97b20c6a7fc637cbf5d780b/src/consumer/multi.rs#L381-L391
- We hit None at https://github.com/streamnative/pulsar-rs/blob/a14e8a15144a48d7d97b20c6a7fc637cbf5d780b/src/consumer/multi.rs#L377-L379

This means we still rip out the topic out of the MultiTopicConsumer forever.

I originally was going to apply a similar fix to #379, but review of the code made me realize that there was code in:

https://github.com/streamnative/pulsar-rs/blob/a14e8a15144a48d7d97b20c6a7fc637cbf5d780b/src/consumer/multi.rs#L145C12-L145C25

That actually handle keeping the initial topics around by recreating them.

It however, wasn't being called because the call was guarded by a topic_regex existence check.

I believe it is necessary for this code to be unconditionally called, because errors in consumers can drop them from a MultiTopicConsumer and we need to recreate.

Local testing confirmed that it works, with 5 full restarts of our staging pulsar cluster while under loading not hanging up once.

### DIFF
--- a/src/consumer/multi.rs
+++ b/src/consumer/multi.rs
@@ -194,7 +194,7 @@ impl<T: DeserializeMessage, Exe: Executor> MultiTopicConsumer<T, Exe> {
                     }),
             )
             .await?;
-            trace!("created {} consumers", consumers.len());
+            info!("created {} consumers", consumers.len());
             Ok(consumers)
         }));
     }
@@ -352,11 +352,9 @@ impl<T: 'static + DeserializeMessage, Exe: Executor> Stream for MultiTopicConsum
             }
         }
 
-        if self.topic_regex.is_some() {
-            if let Poll::Ready(Some(_)) = self.refresh.as_mut().poll_next(cx) {
-                self.update_topics();
-                return self.poll_next(cx);
-            }
+        if let Poll::Ready(Some(_)) = self.refresh.as_mut().poll_next(cx) {
+            self.update_topics();
+            return self.poll_next(cx);
         }
 
         let mut topics_to_remove = Vec::new();
@@ -383,11 +381,7 @@ impl<T: 'static + DeserializeMessage, Exe: Executor> Stream for MultiTopicConsum
                             "Unexpected error consuming from pulsar topic {}: {}",
                             &topic, e
                         );
-                        // Only remove topic from MultiTopicConsumer on error if they
-                        // can be re-added later by regex
-                        if self.topic_regex.is_some() {
-                            topics_to_remove.push(topic.clone());
-                        }
+                        topics_to_remove.push(topic.clone());
                     }
                 }
             } else {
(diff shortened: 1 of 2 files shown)


chamons:
@BewareMyPower - You reviewed my previous PR over this issue. I believe this is the real proper fix.

src/consumer/multi.rs
```diff
@@ -194,7 +194,7 @@ impl<T: DeserializeMessage, Exe: Executor> MultiTopicConsumer<T, Exe> {
                     }),
             )
             .await?;
-            trace!("created {} consumers", consumers.len());
+            info!("created {} consumers", consumers.len());
```
chamons, line 197: We can drop this if needed, but I think it is useful info to have.

chamons:
@mdeltito - FYI

mdeltito:
@chamons Excellent find! This clears up my confusion as well - the removal of topics on fatal errors makes a lot more sense when there is a consistent, out-of-band mechanism for ensuring a connection/consumer exists for all topics.

I suppose this does mean for the (likely much more common) case of _static_ topics, we end up hitting the broker a lot more often to fetch the partitioned topic metadata. Unclear whether that is a "problem" or not; there is not much to go off in the [PR](https://github.com/streamnative/pulsar-rs/pull/327) or [commit](https://github.com/streamnative/pulsar-rs/pull/327/changes/db6d5f77e60179700ca390b102b356ce3fe357ab) that introduced the original change, as I'm sure you saw.

chamons:
That is concerning that it was originally removed for what appears to be performance reasons (?), however if any broker issue can cause the topics to be removed then I think was strictly need to do those refreshes, else you can drop some (or in one case locally, all) topics out of a multi consumer.

A once a 60 second query for all of the topics can't be _that_ bad, right? (Famous last words)

mdeltito:
Totally agree, this behavior and erring the side of being chatty with the broker seems preferable to the current state. 

> A once a 60 second query for all of the topics can't be that bad, right? 

I too want to believe 😄 This also handles the case where a topic for a static subscription has the number of partitions increased. For our use-case we are either forced to use a regex subscription and leverage that refresh interval to pick up the new partition(s), or restart consumers with a static subscription.

chamons:
> has the number of partitions increased

That is a very good point, that can happen at an arbitrary point, is non-error, and we need to handle as well.

BewareMyPower (approved):

BewareMyPower:
FYI, Pulsar Java client has two configs for such periodical tasks:
- `patternAutoDiscoveryPeriod` for pattern consumer to detect changes of topics that match the regex
- `autoUpdatePartitionsInterval` for multi-topics consumer to find increased partitions

Though it seems reasonable to use a single configuration for it

PR #387, opened 2026-01-14
fix: prevent check_connections from dropping connections during setup

### LINKED ISSUE #386: Race Condition in `check_connections` Dropping Active Connections
## Problem
When using `topic_regex` or multi-topic consumers, there is a race condition between the `connect()` function and the `check_connections()` cleanup task that can cause active connections to be prematurely dropped. Now that code has been added to call `update_topics` not only for regex, but for partitioned topics, the race condition will happen frequently for any user of multi-topic consumers.

## Cause
The `check_connections()` function uses `Arc::strong_count()` to determine if a connection is still in use:

```rust
// Original logic
conn.is_valid() && strong_count > 1
```

The issue is that during `connect()`, the Arc is temporarily downgraded to create a weak reference for the keepalive task:

```rust
// In connect()
let c = Arc::new(conn);           // strong_count = 1
let weak_conn = Arc::downgrade(&c);  // still strong_count = 1
// ... spawn keepalive task ...
self.connections.insert(broker, ConnectionStatus::Connected(c.clone()));  // strong_count = 2
```
If `check_connections()` runs between `Arc::downgrade()` and `insert()`, it sees `strong_count == 1` and incorrectly concludes the connection is unused, dropping it from the pool.

## Symptoms
* Repeated "Connected to..." log messages for the same broker
* Flood of connection establishment logs during topic refresh
* Inefficient reconnection on every `update_topics()` call
* Increased latency due to unnecessary reconnections

## Proposed Solution
Track connection creation time and add a grace period before considering a connection for cleanup.

darinspivey:
When using topic_regex or multi-topic consumers, there is a race condition between getting a connection for topic refresh and the check_connections cleanup task. The strong_count check could cause connections to be dropped if check_connections runs while connect() is in progress, since connect() temporarily downgrades the Arc ref.

This commit adds a 5-second grace period for newly created connections, allowing them to be picked up by consumers/producers before being considered for cleanup.

Fixes: #386

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -1,4 +1,8 @@
-use std::{collections::HashMap, sync::Arc, time::Duration};
+use std::{
+    collections::HashMap,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
 use futures::{channel::oneshot, lock::Mutex};
 use rand::Rng;
@@ -150,7 +154,10 @@ impl Default for TlsOptions {
 }
 
 enum ConnectionStatus<Exe: Executor> {
-    Connected(Arc<Connection<Exe>>),
+    Connected {
+        conn: Arc<Connection<Exe>>,
+        created_at: Instant,
+    },
     Connecting(Vec<oneshot::Sender<Result<Arc<Connection<Exe>>, ConnectionError>>>),
 }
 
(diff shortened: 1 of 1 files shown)


darinspivey:
@BewareMyPower you've been helping review some of @chamons contributions recently, and this PR pertains to those (PR #384 ).  Now that `update_topics()` is being called for all multi consumers, the race condition described in this PR will be more frequent.  This provides a fix that I've tested locally. 

Thanks in advance.

BewareMyPower (approved):

chamons (commented):
Fix looks reasonable to me, good catch.

PR #390, opened 2026-01-30
fix(consumer): `update_topics` should only log if topics are recreated

### LINKED ISSUE #389: Logging is too noisy for multi `update_topics`
When `update_topics` is called to periodically make sure all partitions are connected with a consumer, the logging level is at `info!`.  In a production system with lots of topics this creates very noisy logging since it mostly says `created 0 consumers`.

Consider changing this to remain `info!` but only log if it indeed found partitions to be re-connected.  This way, one could search their logs to see when the system is self-healing after consumers drop due to broker/connectivity errors.

https://github.com/streamnative/pulsar-rs/blob/771b6ef81b55577095769736eb47932fe64788bb/src/consumer/multi.rs#L197

darinspivey:
When using `multi` topics (or regex), when disconnected topic partitions are found, it will recreate their consumers. The logging statement for this was quite noisy in a production environment. Change the code to only log if it finds missing partitions, and also log the topic names so that this functionality can be traced in logs.

Fixes: #389

### DIFF
--- a/src/consumer/multi.rs
+++ b/src/consumer/multi.rs
@@ -194,7 +194,14 @@ impl<T: DeserializeMessage, Exe: Executor> MultiTopicConsumer<T, Exe> {
                     }),
             )
             .await?;
-            info!("created {} consumers", consumers.len());
+            if !consumers.is_empty() {
+                let topics: Vec<String> = consumers.iter().map(|c| c.topic()).collect();
+                info!(
+                    "recreated {} consumers for topics: {:?}",
+                    consumers.len(),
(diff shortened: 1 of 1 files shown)


darinspivey:
@BewareMyPower do you mind having a look at this small change?  The `info!` logs are blowing up our logging in production :)

cc: @chamons

chamons (approved):
When I originally info! this I didn't realize how often this would hit. Good catch.

BewareMyPower (approved):

PR #392, opened 2026-02-10
feat: Add an option to control connection cleanup

### LINKED ISSUE #391: `check_connections()` is deleting connections that could be used by `update_topics()`
Previously, some work went into improving the logic around idle connection cleanup. The logic was modified to be more tolerant of a race condition between newly-created connections and the `check_connection()` strong reference counter.

A new bug has surfaced now that `update_topics()` is called periodically with `multi` topics as well as regex topics, both of would benefit from a cached connection.  That update operation will require a connection, and with a topic refresh upwards of 1-2 minutes (it's configurable), this creates a situation where the 5-second grace period does nothing and the connections are deleted.  Once this happens, the topic update interval will constantly create a new connection.  Logs from such an occurrence:

```
Feb 10 14:56:38  INFO pulsar::connection_manager: Removing unused connection d57ab40b-c012-4f0b-ba87-bf63459fd95b to pulsar://pulsar-proxy.pulsar.svc.cluster.local:6650 (strong_count: 1, age: 59.980120242s)
  
Feb 10 14:56:38 INFO pulsar::connection_manager: No existing connection, creating new for pulsar://pulsar-proxy.pulsar.svc.cluster.local:6650
```

In this real-world example, we can see that the timing of `check_connections` fired right before a connection was needed for the `update_topics` interval.  Given that a cached connection will always be beneficial to `update_topics` it might be useful to have a configurable option such that the user can control how much idle time a connection should have before being cleaned up.

darinspivey:
The `check_connections()` function is responsible for cleaning up unused connections. Now that `update_topics` is periodically called for `multi` topics (partitions) as well as for topic regex, it's important that there be a cached connection to use. This commit makes changes to refresh a timestamp when a connection is used and only delete it after a certain configured idle time has passed. A configurable option has been added to give the user full control of the connection cleanup behavior.

Fixes: #391

### DIFF
--- a/src/connection_manager.rs
+++ b/src/connection_manager.rs
@@ -35,6 +35,8 @@ pub struct ConnectionRetryOptions {
     pub connection_timeout: Duration,
     /// keep-alive interval for each broker connection
     pub keep_alive: Duration,
+    /// maximum idle time before a connection is eligible for cleanup
+    pub connection_max_idle: Duration,
 }
 
 impl Default for ConnectionRetryOptions {
@@ -46,6 +48,7 @@ impl Default for ConnectionRetryOptions {
             max_retries: 12u32,
             connection_timeout: Duration::from_secs(10),
             keep_alive: Duration::from_secs(60),
+            connection_max_idle: Duration::from_secs(120),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


darinspivey:
@BewareMyPower this is a follow-up to my recent fix for the connection cleanup race condition.  Thank you!

BewareMyPower (approved):

PR #397, opened 2026-02-27
fix: Preserve message metadata when sending to DLQ

PookieBuns:
Mimic behavior of java client from https://github.com/apache/pulsar/blob/4cbe124841fdc92ab671b1e438a86522a70bd622/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L2345

Also includes #294 since the modifications to the tests require ordering key to be properly forwarded in the Message.

Because of the test changes it should properly address the request in https://github.com/streamnative/pulsar-rs/pull/294#pullrequestreview-1620269534 to have a test that guards against this.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1543,6 +1543,7 @@ pub(crate) mod messages {
                     publish_time: Utc::now().timestamp_millis() as u64,
                     replicated_from: None,
                     partition_key: message.partition_key,
+                    ordering_key: message.ordering_key,
                     replicate_to: message.replicate_to,
                     compression: message.compression,
                     uncompressed_size: message.uncompressed_size,
--- a/src/consumer/batched_message_iterator.rs
+++ b/src/consumer/batched_message_iterator.rs
@@ -51,6 +51,7 @@ impl Iterator for BatchedMessageIterator {
             let metadata = Metadata {
                 properties: batched_message.metadata.properties,
                 partition_key: batched_message.metadata.partition_key,
+                ordering_key: batched_message.metadata.ordering_key,
                 event_time: batched_message.metadata.event_time,
                 ..self.metadata.clone()
             };
--- a/src/consumer/engine.rs
+++ b/src/consumer/engine.rs
@@ -29,6 +29,9 @@ use crate::{
     Error, Executor, Payload, Pulsar,
 };
 
+const SYSTEM_PROPERTY_REAL_TOPIC: &str = "REAL_TOPIC";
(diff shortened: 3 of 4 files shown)


codelipenghui (commented):
Overall, this is a great fix! It correctly preserves the message metadata (, , , and ) when pushing to the DLQ, and fixes the bug where  was not being forwarded in .

A couple of minor suggestions for full parity with the Java client:

1. **Batched Messages Edge Case:** In , we are currently mapping , , and  from the  to individual messages, but we are missing . It would be great to add  to the  struct update there so that individual messages in a batch properly retain or clear their specific ordering keys.

2. **System Properties (Optional / Follow-up):** The Java client injects additional system properties when sending to the DLQ, specifically  (the original topic name) and  (the original message ID). This is extremely helpful for operators inspecting DLQ topics to trace where a message came from. It might be worth adding these to the  map in  before sending to the DLQ, either in this PR or a follow-up.

Otherwise, the code looks solid and the tests cover the new behavior well. LGTM!

codelipenghui (commented):
This is a great fix! It correctly preserves the message metadata (`properties`, `partition_key`, `ordering_key`, and `event_time`) when pushing to the DLQ, and fixes the bug where `ordering_key` was not being properly forwarded in `src/connection.rs`.

A couple of minor suggestions for full parity with the Java client:

1. **Batched Messages Edge Case:** In `src/consumer/batched_message_iterator.rs`, we are currently mapping `properties`, `partition_key`, and `event_time` from the `SingleMessageMetadata` to individual messages, but we are missing `ordering_key`. It would be great to add `ordering_key: batched_message.metadata.ordering_key,` to the `Metadata` struct update there so that individual messages in a batch properly retain or clear their specific ordering keys.

2. **System Properties (Optional / Follow-up):** The Java client injects additional system properties when sending to the DLQ, specifically `REAL_TOPIC` (the original topic name) and `ORIGIN_MESSAGE_ID` (the original message ID). This is extremely helpful for operators inspecting DLQ topics to trace where a message came from. It might be worth adding these to the `properties` map in `src/consumer/engine.rs` before sending to the DLQ, either in this PR or a follow-up.

Otherwise, the code looks solid and the tests cover the new behavior well. LGTM!

codelipenghui (approved):

PookieBuns:
Addressed 1 in a8af8d708bf37711a5929461b7104d54f3faf126

Addressed 2 in 7e924ad8c5549e5b60c68d7bc4a9a104ec0488ff

PR #398, opened 2026-03-12
fix: store and auto-attach schema_version from CommandProducerSuccess

codelipenghui:
## Summary

- The Rust Pulsar client discards the `schema_version` returned by the broker in `CommandProducerSuccess`, so consumers receive messages with no `schema_version` in their metadata — breaking schema evolution and `AutoConsumeSchema` workflows
- Store the broker-returned `schema_version` in `TopicProducer` and auto-attach it to outgoing messages in both the non-batch (`send_raw`) and batch (`message_send_loop`) paths
- On reconnection, update the producer-level `schema_version` for future messages without mutating in-flight messages (whose `schema_version` matches the schema used to serialize their payload)

### `src/retry_op.rs`
- Change `retry_create_producer` return type from `Result<String, Error>` to `Result<(String, Option<Vec<u8>>), Error>` to return `schema_version` alongside `producer_name`
- Capture `schema_version` from the exclusive-access wait path, with warn logging when the broker drops it

### `src/producer.rs`
- Add `schema_version: Option<Vec<u8>>` field to `TopicProducer`
- In `send_raw` (non-batch path): auto-attach `schema_version` when not explicitly set by the user
- In `send_raw` (batch path): document that per-message `schema_version` is structurally dropped (Pulsar's `SingleMessageMetadata` has no such field)
- Pass `schema_version` to `message_send_loop` for the batch envelope
- In `send_message` reconnection: update `*schema_version` for future messages only; never mutate the in-flight `ProducerMessage`
- Add warn-level logging with producer identity and hex-formatted values for all `schema_version` transitions on reconnect (changed, lost, appeared)

### `src/consumer/mod.rs`
- Add shared test helpers (`PULSAR_ADDR`, `init_logger()`, `new_client()`, `schema_producer()`, `earliest_consumer()`)
- `schema_version_set_on_produced_messages` — non-batch producer with JSON schema, verifies `schema_version` is present and non-empty
- `schema_version_set_on_batched_messages` — batched producer with `batch_size=2`, verifies both messages carry the same `schema_version`
- `user_provided_schema_version_not_overwritten` — custom `SerializeMessage` impl sets explicit `schema_version`, verifies it survives round-trip
- `schemaless_producer_has_no_schema_version` — producer without schema, verifies `schema_version` is absent or empty

## Test plan

- [x] `cargo test schema_version` — all 4 new tests pass
- [x] `cargo test` — all existing tests still pass
- [x] `cargo clippy` — no new warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -584,6 +584,19 @@ struct TopicProducer<Exe: Executor> {
     sequence_id: SerialId,
     compression: Option<Compression>,
     options: ProducerOptions,
+    /// Schema version returned by the broker in `CommandProducerSuccess`.
+    ///
+    /// * **Non-batched path** — used as a default when the message does not
+    ///   already have a `schema_version` set (`send_raw`).  Updated in-place
+    ///   on reconnection via `send_message`.
+    /// * **Batched path** — a clone is passed by value to the spawned
+    ///   `message_send_loop`, which owns its own independent copy.  That copy
+    ///   is updated on reconnection within the loop; this field is **not**
+    ///   kept in sync and may go stale for batched producers.
+    ///
+    /// Per-message `schema_version` is not supported by the Pulsar protocol
+    /// for batched messages (`SingleMessageMetadata` has no such field).
+    schema_version: Option<Vec<u8>>,
 }
 
 impl<Exe: Executor> TopicProducer<Exe> {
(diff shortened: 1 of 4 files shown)


src/producer.rs
```diff
@@ -933,6 +964,41 @@ where
...
+            (None, Some(new)) => {
+                warn!(
+                    "schema_version appeared on reconnect for \
+                     producer {:?}({}) on topic {}: None -> {:x?} \
+                     (schema may have been added to the topic)",
+                    producer_name, producer_id, topic, new
+                );
+            }
+            _ => {}
+        }
```
BewareMyPower: This warn log seems not make sense. In Java client, it simply assign the schema version. Could you double confirm it?

codelipenghui: @BewareMyPower removed.

PR #399, opened 2026-03-13
feat: support external schema registry (PIP-420)

codelipenghui:
## Summary

- Implement [PIP-420](https://github.com/apache/pulsar/blob/master/pip/pip-420.md) external schema registry support for pulsar-rs
- Add `PulsarSchema<T>` async trait as the unified schema interface for encode/decode with external registries
- Wire `schema_id` propagation end-to-end: `PulsarSchema::encode()` → `EncodeData` → `ProducerMessage` → `MessageMetadata` → wire → consumer decode
- Add `DefaultPulsarSchema<T>` bridging existing `SerializeMessage`/`DeserializeMessage` traits for backward compatibility
- Add `KeyValueSchema<K, V>` composing two inner schemas with magic-byte framing (`0xFE`)
- Add `pulsar-schema-confluent` workspace crate as a Confluent Schema Registry reference implementation

## Changes

**Protocol & Core** (3 commits)
- Sync `PulsarApi.proto` from upstream with `SchemaType::External = 22` and `schema_id` in `MessageMetadata`
- Add `SchemaRegistry` error variant
- Add `schema_id_util` module with magic-byte framing helpers

**Schema Framework** (4 commits)
- `PulsarSchema<T>` trait: `schema_info()`, `encode()`, `decode()`, `close()`
- `EncodeData` struct with `payload` and optional `schema_id`
- `DefaultPulsarSchema<T>` and `KeyValueSchema<K, V>` implementations
- Re-exports from `pulsar::schema`

**Consumer Integration** (4 commits)
- `Message<T>` gains `decoded: Option<T>` field with `value()` accessor
- `DecodedMessageReceiver<T>` type alias for schema-decoded channel
- `ConsumerBuilder::with_schema()` for type-erased schema injection
- Async decode task spawned between `ConsumerEngine` and `TopicConsumer`

**Producer Integration** (1 commit)
- `ProducerBuilder::with_schema()` and `Producer::send_schema_non_blocking()`
- `schema_id` propagated through batching pipeline (`BatchItem` 3-tuple)
- `connection.rs` wires `schema_id` into `MessageMetadata`

**Confluent Reference Implementation** (1 commit)
- `pulsar-schema-confluent` workspace crate with `ConfluentJsonSchema<T>`
- `ConfluentConfig`, `ConfluentAuth`, `SubjectNameStrategy` types
- `ConfluentSchemaFactory` convenience builder
- Schema registration is placeholder for initial release

## Test plan

- [x] All 53 integration tests pass with live Pulsar broker
- [x] All 17 doc-tests pass
- [x] 5 confluent crate unit tests pass (1 ignored — requires live schema registry)
- [ ] Manual test with Confluent Schema Registry (future iteration)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -68,6 +68,7 @@ impl Future for SendFuture {
 /// this is actually a subset of the fields of a message, because batching,
 /// compression and encryption should be handled by the producer
 #[derive(Debug, Clone, Default)]
+#[non_exhaustive]
 pub struct Message {
     /// serialized data
     pub payload: Vec<u8>,
@@ -90,6 +91,17 @@ pub struct Message {
     /// Schema ID from external schema registry (PIP-420).
     /// Written to MessageMetadata.schema_id when present.
     pub schema_id: Option<Vec<u8>>,
+}
+
+impl Message {
+    /// Create a `Message` carrying the given payload with all other fields
+    /// set to their defaults.
+    pub fn new(payload: Vec<u8>) -> Self {
+        Self {
+            payload,
+            ..Default::default()
+        }
+    }
 }
 
 /// internal message type carrying options that must be defined
@@ -1091,7 +1103,8 @@ impl<Exe: Executor> Clone for ProducerBuilder<Exe> {
             name: self.name.clone(),
             producer_options: self.producer_options.clone(),
             schema_info: self.schema_info.clone(),
-            // schema_object is NOT cloned — only the original builder carries the schema.
+            // schema_object cannot be cloned (Box<dyn Any>). The original builder
+            // retains it; clones are only valid for non-schema paths.
             schema_object: None,
         }
     }
@@ -1373,14 +1386,45 @@ async fn message_send_loop<Exe>(
                 };
                 let counter = batch_items.len();
 
-                // Extract schema_id from first message (PIP-420: all messages in batch share same schema_id)
-                let batch_schema_id = batch_items.iter().find_map(|item| {
-                    if let BatchItem::SingleMessage(_, _, schema_id) = item {
-                        schema_id.clone()
-                    } else {
-                        None
+                // Extract and validate schema_id across the batch (PIP-420).
+                // All messages must share the same schema_id; mixed batches are rejected.
+                let batch_schema_id = {
+                    let mut first_id: Option<Option<Vec<u8>>> = None;
+                    let mut conflict = false;
+                    for item in batch_items.iter() {
+                        if let BatchItem::SingleMessage(_, _, schema_id) = item {
+                            match &first_id {
+                                None => first_id = Some(schema_id.clone()),
+                                Some(prev) if prev != schema_id => {
+                                    conflict = true;
+                                    break;
+                                }
+                                _ => {}
+                            }
+                        }
                     }
-                });
+                    if conflict {
+                        error!(
+                            "producer {}: batch contains messages with different schema_ids — \
+                             rejecting entire batch",
+                            producer_id
+                        );
+                        // Fail all messages in the batch
+                        for item in batch_items {
+                            if let BatchItem::SingleMessage(tx, _, _) = item {
+                                let _ = tx.send(Err(Error::Custom(
+                                    "Batch rejected: messages have conflicting schema_ids"
+                                        .to_string(),
+                                )));
+                            }
+                        }
+                        if let Some(flush_tx) = flush_tx {
+                            let _ = flush_tx.send(());
+                        }
+                        continue;
+                    }
+                    first_id.flatten()
+                };
 
                 for item in batch_items {
                     if let BatchItem::SingleMessage(tx, batched_msg, _) = item {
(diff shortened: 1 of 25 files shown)


BewareMyPower:
You might need to modify the AGENTS.md like:

> Don't modify existing code unless necessary

src/producer.rs
```diff
@@ -197,6 +210,11 @@ pub struct MultiTopicProducer<Exe: Executor> {
     producers: BTreeMap<String, Producer<Exe>>,
     options: ProducerOptions,
     name: Option<String>,
+    /// Type-erased schema object, moved from `ProducerBuilder::with_schema`.
+    /// Used by [`send_schema_non_blocking`](Self::send_schema_non_blocking) to
+    /// encode messages at the multi-topic level before delegating to per-topic
+    /// producers.
+    schema_object: Option<Box<dyn Any + Send + Sync>>,
```
PookieBuns, line 217: Hi @codelipenghui, may I provide a recommendation to a separate approach than storing a `Box<dyn Any + Send + Sync>`?

The current implementation has a drawback in which types that aren't supported by the `Schema` are not invalidated at compile time but instead will produce a runtime error everytime `send_schema_non_blocking` is called when it tries to downcast into the `Schema` object. This is not ideal since if we stored the actual trait object we would have compile time guaranteed correctness of schema supported T's. 

Assuming we would not want to make any breaking changes, and can't just force Producer to suddenly require type annotations, I have 2 recommendations that we could potentially take.

1. Introduce a new type of producer like `SchemaProducer`, which owns the actual trait object and gets its own version of `send_non_blocking`. It would be initialized with something like `Producer::with_schema`  which returns a `SchemaProducerBuilder`, or we could have the `with_schema` method on the `ProducerBuilder` create a type change.

The downside of this approach is the cardinality, since then there would be 3 extra structs to maintain (`SchemaProducer`,`SchemaProducerBuilder`,`MultitopicSchemaProducer`, and potentially `TypedMessageBuilder`).

2. Add a generic type parameter T which defaults to `Vec<u8>` to the producer type, and the default producer, will hold `<Box<dyn Schema<Vec<u8> + Send + Sync>>`. The default_schema would only implement `Schema` for `Vec<u8>` (or introduce a bytes_schema). Then, send_schema by default will only take in `Vec<u8>`'s. 

This is actually more similar to the java approach since initializing a producer without a schema will default to bytes schema.

However, I'm not 100% sure if this is non-breaking since it technically attaches a generic to the existing `Producer` (although with a default)

src/producer.rs
```diff
@@ -434,6 +522,100 @@ impl<Exe: Executor> Producer<Exe> {
...
+    /// parameter does not match `T`, an error is returned.
+    ///
+    /// Use this method (instead of [`send_non_blocking`]) when you want
+    /// schema-aware encoding for messages produced with an external schema
+    /// registry.
+    ///
+    /// The `T: 'static` bound is required for Rust's `Any`-based runtime
+    /// type-checking inside the type-erased schema dispatch.
+    #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    pub async fn send_schema_non_blocking<T: SerializeMessage + Sized + Send + 'static>(
```
PookieBuns, line 539: Can we also introduce a `create_schema_message` or `create_message_with_schema` that gets something like a `TypedMessageBuilder`, parallel to `MessageBuilder` which allows building a message that utilizes the schema? This is similar to `TypedMessageBuilder` in java

src/schema/key_value.rs
```diff
@@ -0,0 +1,375 @@
...
+use crate::{
+    message::proto,
+    schema::{schema_id_util, EncodeData, PulsarSchema},
+    Error, Payload,
+};
+
+use super::schema_id_util::SchemaIdInfo;
+
+/// Composes two inner schemas for key-value message types.
+pub struct KeyValueSchema<K, V> {
```
PookieBuns, line 14: Is it possible to support [SEPARATED](https://pulsar.apache.org/docs/next/schema-understand/#keyvalue-schema) ? Java implements this by doing a [setSeparateKeyValue](https://github.com/apache/pulsar/blob/ce5bc1364b108dfe69fadbaf9d838a44b17aa2ae/pulsar-client/src/main/java/org/apache/pulsar/client/impl/TypedMessageBuilderImpl.java#L81)

src/producer.rs
codelipenghui: The primary reason is compatibility. The first option seems like a viable solution that avoids breaking changes, though it would increase the maintenance burden. I will give it some thought.

src/producer.rs
codelipenghui: I can follow up with a separate PR, keep this PR to focus on the schema abstraction and external schema support.

src/schema/key_value.rs
codelipenghui: I can follow up with a separate PR, keep this PR to focus on the schema abstraction and external schema support.

src/producer.rs
BewareMyPower: This comment makes sense. I think the main concern is it introduces a separated public API (`send_schema_non_blocking`) to send messages. However, like other client implementation, we should reuse the existing `send_non_blocking` method to send a message with custom schema.

BewareMyPower (changes requested):

PR #401, opened 2026-03-22
Fix producer not reconnecting after receipt timeout

### LINKED ISSUE #400: Producer stuck in infinite timeout loop when broker silently blocks sends
When the broker does not reply to `CommandSend` (e.g. due to backlog quota with
`block_if_queue_full`), the producer enters an infinite timeout loop:

1. `send_message` enqueues the message via `sender().send()` — succeeds (channel is open)
2. The receipt future times out after `operation_timeout` (default 30s)
3. The timeout poisons the shared connection via `error.set()`
4. On the next call, `sender().send()` still succeeds (the `async_channel` and TCP
   connection are still alive) — same cycle repeats indefinitely

The producer has reconnection logic (`producer.rs:942-973`), but it only triggers when
`sender().send()` fails immediately. When the send succeeds but the receipt times out,
reconnection is never reached.

Note: `ConsumerEngine` handles this case correctly — it checks `is_valid()` and
reconnects. The producer does not.

**Reproduction:** Any scenario where the broker holds `CommandSend` responses beyond
`operation_timeout` — backlog quota with `block_if_queue_full` being the most common case.

---

_Investigation and fix co-authored with [Claude Code](https://claude.com/claude-code)._

CAT-Solstice:
Fixes #400

## Summary

- Check `connection.is_valid()` at the top of the `send_message` loop. If the connection
  is poisoned, fall through to existing reconnection logic instead of sending into a dead
  connection.
- Add `ProducerBlockedQuotaExceededError` and `ProducerBlockedQuotaExceededException`
  to retryable errors in `handle_retry_error`, so producer recreation retries with
  backoff when the broker rejects due to quota.

## Context

When the broker silently blocks sends (e.g. backlog quota with `block_if_queue_full`),
the producer enters an infinite timeout loop because `sender().send()` succeeds
(the channel is still open) but the receipt never arrives. The existing reconnection
logic only triggers on immediate send failure, never on receipt timeout.

`ConsumerEngine` already handles this correctly by checking `is_valid()` — this PR
applies the same pattern to the producer.

## Test plan

- Tested against a Pulsar 2.6.4 broker with backlog quota and `block_if_queue_full` enabled
- Producer now detects poisoned connection, reconnects, and retries creation with
  backoff until quota is freed
- When quota is released, the producer recreates successfully and resumes sending

---

_Co-authored with [Claude Code](https://claude.com/claude-code)._

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -905,39 +905,52 @@ where
     Exe: Executor,
 {
     loop {
-        match connection
-            .sender()
-            .send(
-                producer_id,
-                producer_name.clone(),
-                sequence_id.get(),
-                message.clone(),
-                options.block_queue_if_full,
-            )
-            .await
-        {
-            Ok(fut) => {
-                let fut = async move {
-                    let res = fut.await;
-                    res.map_err(|e| {
-                        error!("wait send receipt got error: {:?}", e);
-                        Error::Producer(ProducerError::Connection(e))
-                    })
-                };
-                return Ok(fut);
-            }
-            Err(ConnectionError::Disconnected) => {}
-            Err(ConnectionError::Io(e)) => {
-                if e.kind() != std::io::ErrorKind::TimedOut {
-                    error!("send_message got io error: {:?}", e);
-                    return Err(ProducerError::Connection(ConnectionError::Io(e)).into());
+        // If a previous send timed out waiting for a receipt, the connection
+        // is poisoned (error flag set) but the underlying TCP channel may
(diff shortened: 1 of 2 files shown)


chamons:
We _just_ ran into a producer hang after broker restart last week and I was looking into it when I came across this PR.

@BewareMyPower - Can we get a review? This looks great

I am working locally to see if I can get a repro to test this, but since you need a delayed receipt that's hard to line up.

BewareMyPower (approved):

PR #403, opened 2026-03-25
feat: add Pulsar Admin REST API client for topic policy management

### LINKED ISSUE #143: HTTP client for the Pulsar API
should we provide a client for the Pulsar HTTP API, through which we can manage subscriptions, get statistics, etc?

### LINKED ISSUE #402: Need admin API client for topic policy management
## Motivation

In a pulsar cluster where auto topic creation is enabled, it's difficult to automate the creation of any topic-level policies since topics must obviously exist before policies can be set.  The most guaranteed place to set such policies is in the consumer, which will create topics that don't exist.  This applies to consumers who are restarted against topics (or partitions) that have been auto-deleted via GC policies as well, and they'll be re-created allowing for topic policies to be set.

A powerful Pulsar feature that can be controlled with an admin api client is the ability to set "max unacked messages per consumer" (and possibly subscription) to provide backpressure if clients are failing to ack properly.

## Suggestion

Scaffold out a configuration for broker-level policies to be set via an admin API call when consumers start.  This could be worked into the builder methods and support things like `max_unacked_messages_per_consumer` to start, but leaving room for other settings to be added later.  Optionally, users could also call their own apis by accessing the exposed admin http client directly.

darinspivey:
Adds optional `admin-api` feature backed by reqwest/native-tls that exposes an AdminClient for setting broker-side topic policies (e.g. maxUnackedMessagesOnConsumer). Policies can be applied manually via pulsar.admin() and its associated methods. Updates CI to enable the feature and set topicLevelPoliciesEnabled=true on the test broker. Also, drop Pulsar 2.10.6 from test matrix because that version does not support topicLevelPoliciesEnabled via Docker env vars. The code will still work on 2.10.6; this is just a test concern.

The code for TLS and other authentication bits was done in the same way as the auth in the rest of this client create.

Fixes: #402
Fixes: #143

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -1,0 +1,259 @@
+//! Pulsar Admin REST API client.
+//!
+//! Enabled by the `admin-api` feature flag. Requires a tokio runtime.
+
+use std::{sync::Arc, time::Duration};
+
+use futures::lock::Mutex;
+
+use crate::{
+    authentication::Authentication,
+    connection_manager::TlsOptions,
+    error::{AdminError, Error},
+};
+
+/// Parses a Pulsar topic URL into (scheme, tenant, namespace, topic_name).
+/// Accepts `persistent://` and `non-persistent://` prefixes, or a bare
+/// `tenant/namespace/topic` string which defaults to `persistent://`.
+fn parse_topic(topic: &str) -> Result<(&str, &str, &str, &str), Error> {
+    let invalid = || {
+        Error::Admin(AdminError::InvalidTopic(format!(
+            "expected tenant/namespace/topic or a fully-qualified topic URL, got: {topic}"
+        )))
+    };
+
+    let (scheme, rest) = if let Some(rest) = topic.strip_prefix("persistent://") {
+        ("persistent", rest)
+    } else if let Some(rest) = topic.strip_prefix("non-persistent://") {
+        ("non-persistent", rest)
+    } else {
+        ("persistent", topic)
+    };
+
+    let mut parts = rest.splitn(3, '/');
+    let tenant = parts.next().filter(|s| !s.is_empty()).ok_or_else(invalid)?;
+    let namespace = parts.next().filter(|s| !s.is_empty()).ok_or_else(invalid)?;
+    let name = parts.next().filter(|s| !s.is_empty()).ok_or_else(invalid)?;
+
+    Ok((scheme, tenant, namespace, name))
+}
+
+/// Client for the Pulsar Admin REST API.
+///
+/// Obtain an instance via [`Pulsar::admin()`][crate::Pulsar::admin].
+///
+/// # Example
+///
+/// ```rust,no_run
+/// # async fn run(pulsar: pulsar::Pulsar<pulsar::TokioExecutor>) -> Result<(), pulsar::Error> {
+/// let admin = pulsar.admin("http://localhost:8080")?;
+/// admin
+///     .set_max_unacked_messages_per_consumer(
+///         "persistent://public/default/my-topic",
+///         500,
+///     )
+///     .await?;
+/// # Ok(())
+/// # }
+/// ```
+pub struct AdminClient {
+    client: reqwest::Client,
+    admin_url: String,
+    auth: Option<Arc<Mutex<Box<dyn Authentication>>>>,
+}
+
+impl AdminClient {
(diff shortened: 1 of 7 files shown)


darinspivey:
@BewareMyPower we've been using this on our Pulsar clusters for several weeks and it has helped control our consumer backpressure.  We use auto topic creation, so this allows us to set topic-level policies in an automated fashion.

Please consider this addition, as it appears others have requested http api access before.

src/consumer/builder.rs
```diff
@@ -277,6 +306,29 @@ impl<Exe: Executor> ConsumerBuilder<Exe> {
...
+        {
+            let admin = crate::admin::AdminClient::new(
+                admin_url.clone(),
+                &self.pulsar.manager.tls_options,
+                self.pulsar.manager.auth.clone(),
+            )?;
+            for topic in topics {
+                if let Some(max_unacked) = broker_config.max_unacked_messages_per_consumer {
+                    admin
+                        .set_max_unacked_messages_per_consumer(topic, max_unacked)
```
BewareMyPower: This should not be set from client side. Coupling a consumer with the Pulsar admin also looks weird.

BewareMyPower (changes requested):

src/consumer/builder.rs
darinspivey: Yes, I understand.

Would this make more sense if **only** an admin client was exposed for use?  I feel like `pulsar-rs` should offer some sort of ability to connect to the apis to do topic-level things.  If I'm being honest, setting things like "max unacked" is really a client concern since it's directly related to the consumption of the data.  The fact that pulsar has this labeled as an "admin api" feels unwarranted.

It seems like other people have been asking for something like this over time.  It would be nice if `pulsar-rs` could at least offer the client for use with such things.  Again, for topics that are auto-created, there's no better place than to have the pulsar client be able to manage such topic policies after they've been auto-created and guaranteed to exist.

Would you accept this PR if it just exposed the admin client for arbitrary use?

BewareMyPower: Yes, a separated admin client is okay. Actually, to achieve the same goal, you can manually combine a consumer and an admin (to update `max_unacked_messages_per_consumer`), rather than embedding the admin operation into a consumer

darinspivey: I have modified the code to decouple the use of the admin client from the consumer.  I will test this in my fork prior to re-requesting review.  Thanks!

src/client.rs
```diff
@@ -561,6 +600,20 @@ impl<Exe: Executor> PulsarBuilder<Exe> {
         self
     }
 
+    /// Sets the base URL of the Pulsar Admin HTTP endpoint (e.g. `"http://localhost:8080"`).
+    ///
+    /// When set, consumer builders can automatically apply broker-side topic
+    /// policies (such as [`with_max_unacked_messages_per_consumer`][crate::ConsumerBuilder::with_max_unacked_messages_per_consumer])
+    /// immediately after subscription.
```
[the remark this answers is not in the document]
darinspivey: Closing these comments as they're outdated due to code changes.

src/admin.rs
```diff
@@ -0,0 +1,257 @@
...
+            }
+        }
+
+        Ok(AdminClient {
+            client: builder
+                .build()
+                .map_err(|e| Error::Admin(AdminError::Request(e)))?,
+            admin_url: admin_url.trim_end_matches('/').to_string(),
+            auth,
+        })
```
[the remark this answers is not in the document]
darinspivey: fixed

src/admin.rs
```diff
@@ -0,0 +1,257 @@
...
+        let mut builder = reqwest::ClientBuilder::new()
+            .danger_accept_invalid_certs(tls_options.allow_insecure_connection);
+
+        builder = builder.danger_accept_invalid_hostnames(
+            tls_options.allow_insecure_connection || !tls_options.tls_hostname_verification_enabled,
+        );
+
+        if let Some(pem_bytes) = &tls_options.certificate_chain {
+            let certs = pem::parse_many(pem_bytes)
+                .map_err(|e| Error::Custom(format!("failed to parse certificate chain: {e}")))?;
```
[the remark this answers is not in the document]
darinspivey: fixed

BewareMyPower (approved):

PR #419, opened 2026-06-25
feat: add negative ack redelivery delay and backoff support (PIP-106)

abbudao:
## Summary

Adds negative-acknowledgment redelivery delay and backoff support, implementing [PIP-106](https://github.com/apache/pulsar/blob/master/pip/pip-106.md).

 Before this change, Rust `nack` effectively requested immediate broker redelivery. Other Pulsar clients already support delayed negative-ack redelivery, and Go/Java default to a 60s fixed delay. This PR brings the Rust client closer to those established semantics while keeping the public API idiomatic for this crate.

This contribution was tested in production at my company for two weeks and so far we haven't noticed any issues.

 ## ⚠ Breaking behavior change

 **Default negative-ack redelivery delay changed from 0 ms to 60 seconds** (Go/Java parity).
 Consumers that relied on immediate redelivery on negative ack will now wait 60 seconds
 before the message is redelivered.

 **Suggested release note:**
 > Negative acknowledgments now delay redelivery by 60 s by default (Go/Java parity) instead
 > of requesting immediate redelivery. Use `.with_nack_redelivery_delay(Duration::ZERO)` to
 > restore the previous behavior.

 **Semver implication:** this changes observable consumer behavior for any caller using
 `Consumer::nack` or `Consumer::nack_with_id` without configuring a custom delay — treat as
 a minor breaking change for crates.io semver purposes.

 ## Design references

 The design was checked against the existing client implementations:

 - Go: [`Consumer.Nack` / `NackID`](https://github.com/apache/pulsar-client-go/blob/master/pulsar/consumer_impl.go#L678-L705), [`negativeAcksTracker`](https://github.com/apache/pulsar-client-go/blob/master/pulsar/negative_acks_tracker.go#L49-L140), and `ConsumerOptions.NackRedeliveryDelay` / `NackBackoffPolicy`.
 - Java: [`ConsumerImpl.negativeAcknowledge`](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L854-L868), [`NegativeAcksTracker`](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/NegativeAcksTracker.java#L43-L140), and `negativeAckRedeliveryDelay(...)` / `negativeAckRedeliveryBackoff(...)`.

 The Rust API mirrors the same concepts with builder methods and `Duration` values rather than copying either client API directly.

 ## User-facing changes

 - Adds fixed negative-ack redelivery delay configuration with `ConsumerBuilder::with_nack_redelivery_delay(Duration)`.
 - Adds configurable `NegativeAckBackoff` / `MultiplierRedeliveryBackoff` support for message-based negative acknowledgments.
 - Carries broker `redelivery_count` into `Message::redelivery_count()` and internal negative-ack scheduling.
 - Delays negative-ack redelivery client-side through the consumer engine tracker/ticker.
 - Documents and demonstrates the behavior in `README.md` and `examples/negative_ack.rs`.

 ## Implementation notes for reviewers

 Backoff is only applied to message-based nacks:

 ```rust
 consumer.nack(&msg).await?;
 ```

 That path has access to the broker redelivery count, so it can compute the configured backoff delay. ID-only nacks do not carry that count:

 ```rust
 consumer.nack_with_id(message_id).await?;
 ```

 For ID-only nacks, the client falls back to the configured fixed delay, or the 60s default. This follows the same practical limitation called out in the Go client.

 The internal channel types were split deliberately. The broker redelivery count enters through `CommandMessage`, then needs to travel from `ConsumerEngine` to `TopicConsumer` so `Message::redelivery_count()` and message-based nack backoff can work. Changing the existing public-facing `MessageIdDataReceiver` alias from `(MessageIdData, Payload)` to `(MessageIdData, Payload, Option<u32>)` would have been a source-breaking tuple-shape change, so the wider tuple is kept behind `pub(crate)` as `InternalMessageIdDataResult`.

 The same applies to engine messages. `EngineMessage::Nack(MessageIdData)` is still the public shape; the new `InternalEngineMessage::Nack(MessageIdData, Option<u32>)` carries redelivery count only inside the consumer implementation. `InternalEngineEvent` also keeps the private negative-ack timer event out of the public enum.

 The delayed redelivery state is tracked client-side and dispatched by asking the broker to redeliver due message ids. The tracker normalizes batch-specific ids to the broker redelivery entry while still tracking individual pending batch indexes so later acks can cancel the right scheduled nacks.

 A normal ack or cumulative ack cancels matching pending negative-ack redelivery state. This avoids redelivering a message that was nacked and then successfully processed before the delayed redelivery request fired.

 ## Out of scope

 This PR deliberately does not change ack-timeout backoff, DLQ/retry-letter semantics, regex topic matching, reader behavior, or broad consumer lifecycle behavior.


abbudao:
I'd love to understand the project's appetite for PRs like this that fill in missing PIPs. Is this the kind of contribution the maintainers are interested in taking, or is  review bandwidth for larger features tight right now? Genuinely fine either way. I just want to calibrate expectations.

For context: we run pulsar-rs in production and depend on it fairly heavily, so features  like this (plus a couple of batch-consuming bugs we've hit) matter a lot to us. We'd like to contribute more regularly, and knowing the current review capacity, and what you'd  want from external contributors, would help me invest the right way. A PR that stays compatible and is easy to review takes real effort to put together, and I'm happy to keep doing that if there's a path to getting it merged.

cc @jiangpengcheng sorry to tag you directly. You seem to be active on the repo lately, so hopefully you can point me in the right direction (or to  whoever's best placed to weigh in). No rush at all.

PR #424, opened 2026-08-18
feat: Add Java-compatible partition key bytes

ltseeley:
Wire partition_key_b64_encoded through produce, consume, and DLQ so binary keys match the Java keyBytes/getKeyBytes contract.

### DIFF
--- a/src/consumer/message.rs
+++ b/src/consumer/message.rs
@@ -1,9 +1,11 @@
 use std::marker::PhantomData;
 
+use base64::Engine;
+
 use crate::{
     consumer::data::MessageData,
     message::proto::{MessageIdData, MessageMetadata},
-    DeserializeMessage, Payload,
+    DeserializeMessage, Error, Payload,
 };
 
 /// a message received by a consumer
@@ -42,10 +44,41 @@ impl<T> Message<T> {
         &self.message_id.id
     }
 
-    /// Get message key (partition key)
+    /// Get message key (partition key string as stored in metadata)
+    ///
+    /// When [`has_base64_encoded_key`](Self::has_base64_encoded_key) is true, this is the
+    /// Base64 encoding of the raw key bytes.
     #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
     pub fn key(&self) -> Option<String> {
         self.payload.metadata.partition_key.clone()
+    }
+
+    /// Whether the partition key is a Base64 encoding of raw key bytes.
+    #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    pub fn has_base64_encoded_key(&self) -> bool {
+        self.payload.metadata.partition_key_b64_encoded()
+    }
+
+    /// Get the raw partition key bytes.
+    ///
+    /// If the key is Base64-encoded, it is decoded; otherwise the UTF-8 bytes of
+    /// the key string are returned. Returns `Ok(None)` when there is no key.
+    ///
+    /// Returns an error when the key is marked Base64-encoded but decoding fails.
+    #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    pub fn key_bytes(&self) -> Result<Option<Vec<u8>>, Error> {
+        match self.payload.metadata.partition_key.as_ref() {
+            None => Ok(None),
+            Some(key) if !self.has_base64_encoded_key() => Ok(Some(key.as_bytes().to_vec())),
+            Some(key) => match base64::engine::general_purpose::STANDARD.decode(key) {
+                Ok(bytes) => Ok(Some(bytes)),
+                Err(e) => Err(Error::Custom(format!(
+                    "failed to decode Base64 partition key on topic {} message_id={:?}: {e}",
+                    self.topic,
+                    self.message_id()
+                ))),
+            },
+        }
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/consumer/message.rs
```diff
@@ -42,11 +45,42 @@ impl<T> Message<T> {
...
+        self.payload.metadata.partition_key_b64_encoded()
+    }
+
+    /// Get the raw partition key bytes (Java `getKeyBytes()`).
+    ///
+    /// If the key is Base64-encoded, it is decoded; otherwise the UTF-8 bytes of
+    /// the key string are returned. Returns `None` when there is no key, or when
+    /// Base64 decoding fails.
+    #[cfg_attr(feature = "telemetry", tracing::instrument(skip_all))]
+    pub fn key_bytes(&self) -> Option<Vec<u8>> {
```
codelipenghui: `key_bytes()` returns `None` both when there is no key and when the key fails to decode, with only a `debug!` separating them. Java's `getKeyBytes()` throws on malformed Base64 (`MessageImpl.java:688-696`), and the broker logs at ERROR before degrading to `NONE_KEY` — this makes the Rust client the quietest link in the chain, and a silently dropped key on a Key_Shared topic surfaces far from its cause.

Worth settling now: the signature is new, so changing it is free today and a breaking change once it ships. Suggest `Result<Option<Vec<u8>>, Error>`, or keep `key_bytes()` and add `try_key_bytes()`.

If you'd rather keep `Option`, please raise the log to `warn!` and include the topic and message id — as written it names neither, so in a multi-topic consumer you only learn that something somewhere had a bad key.

src/producer.rs
```diff
@@ -1555,6 +1582,124 @@ mod tests {
         assert!(pm.uncompressed_size.is_none());
     }
 
+    #[test]
+    fn with_key_bytes_sets_base64_partition_key_and_flag() {
```
codelipenghui: Neither this test nor `with_key_clears_base64_flag` below calls the method it's named after. Both hand-build a `Message` literal with the expected values already filled in, then assert that `From<Message> for ProducerMessage` copies them — which `message_converts_into_producer_message` already covers at :1572.

So the two pieces of logic this PR actually adds to the producer have no offline coverage: `with_key_bytes`'s encoding is exercised only by the live-broker test, and the `Some(false)` writes in `with_key`/`with_partition_key` aren't covered at all — delete either line and the suite stays green.

`MessageBuilder` needs a `&mut Producer`, so testing it directly is awkward; extracting the encoding into a small `pub(crate) fn` would make it reachable. Failing that, please rename these to what they assert so the names don't imply coverage that isn't there.

(Minor: `with_key_clears_base64_flag` is a misnomer — `with_key` sets the flag to `Some(false)`, it doesn't clear it to `None`.)

src/producer.rs
```diff
@@ -1536,13 +1561,15 @@ mod tests {
...
             deliver_at_time: Some(123456789),
+            ..Default::default()
         };
 
         let pm: ProducerMessage = m.clone().into();
 
         assert_eq!(pm.payload, m.payload);
         assert_eq!(pm.properties, m.properties);
         assert_eq!(pm.partition_key, m.partition_key);
+        assert_eq!(pm.partition_key_b64_encoded, m.partition_key_b64_encoded);
```
codelipenghui: This assertion can't fail. The literal above ends in `..Default::default()`, so both sides are `None`; and since the `From` impl also ends in `..Default::default()`, deleting the mapping at :135 still passes.

Setting `partition_key_b64_encoded: Some(true)` in the literal makes it a real assertion.

codelipenghui (commented):
Went through the produce / consume / batch / DLQ paths against the Java client. All four propagation sites are correct, and the `batched_message_iterator.rs` change quietly fixes a real pre-existing bug — the b64 flag used to leak from the batch envelope onto every message in the batch via `..self.metadata.clone()`, so a mixed batch from a Java producer decoded wrong. That deserves a line in the commit message; right now it reads like plumbing.

Four things I think should land in this PR. Everything else I found is either pre-existing or additive later.

One more, no code change needed: adding a `pub` field to `producer::Message` (not `#[non_exhaustive]`) breaks downstream exhaustive struct literals — the diff had to add `..Default::default()` at producer.rs:1564 for exactly that reason. The next release needs to be 7.0.0.

src/producer.rs
```diff
@@ -1555,6 +1582,124 @@ mod tests {
...
+            .create_message()
+            .with_content(&data)
+            .with_key_bytes(raw_key)
+            .send_non_blocking()
+            .await
+            .unwrap()
+            .await
+            .unwrap();
+
+        let msg = consumer.next().await.unwrap().unwrap();
```
codelipenghui: No timeout here, while every other consuming test uses `recv_within(.., DEFAULT_RECV_TIMEOUT)` — including the new DLQ test in this PR.

CI runs `RUST_TEST_THREADS: 1` with a 30-minute job budget across 6 Pulsar versions, so one hang burns an entire matrix leg and reports nothing useful. `recv_within` is private to the `consumer` test module, so the simplest fix here is wrapping in `tokio::time::timeout`.

src/consumer/message.rs
ltseeley: Updated to return a `Result` as suggested

src/producer.rs
ltseeley: Addressed!

src/producer.rs
ltseeley: Addressed!

src/producer.rs
ltseeley: Updated

codelipenghui (approved):

